PR #13, opened 2020-07-07
Update data_life_cycle.md

OlivierCollin:
Tentative list.

### DIFF
--- a/pages/data_life_cycle.md
+++ b/pages/data_life_cycle.md
@@ -1,12 +1,38 @@
 ---
-title: "This is a test page"
-keywords: 
+title: "First draft"
+keywords: Data Life Cycle 
 sidebar: main
 permalink: data_life_cycle.html
-summary: This is a test summary
+summary: Overview of the Data Life Cycle 
 ---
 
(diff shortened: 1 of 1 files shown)


martin-nc:
Could we change 'Share and publish' to just 'Share'? Then it would be one word like the rest, and be consistent with other RDM diagrams e.g. https://www.fws.gov/data/life-cycle.

OlivierCollin:
Yes, that makes sense for me. 
Does any scientific information specialist has any ideas about this ?

korbinib:
I feel that there is strong feelings connected to the titles of the life cycle stages.
The most wide spread seem to be:
- Re-using data
- Creating data
- Processing data
- Analysing data
- Preserving data
- Giving access to data

The DSW uses some slightly different terms:
- Re-using data
- Creating and collecting data
- Processing data
- Interpreting data
- Preserving data
- Giving access to data

bedroesb:
We went for a diagram on the landing page to refer to the life cycle. Thank you for starting this topic in this PR.

PR #22, opened 2020-07-07
Update functional_areas.md

OlivierCollin:
Description and list of functional areas


bedroesb:
We use the landing page to go to the different functional areas, so I think for now this page is obsolete...

OlivierCollin:
Yes I agree, we can close this PR.

Issue #26, opened 2020-07-08
Inconsistent fonts


pinarpink:
The font used for the top left bookmark "ELIXIR RDM Toolkit" is recognisably different from the other fonts used on the site.

bedroesb:
Working on it ;)

Issue #23, opened 2020-07-07
Ensure a consistent colour scheme for the site


martin-nc:
Change the current template to ELIXIR colours? Most of the template should remain white so I'm not suggesting a frenzy of orange. The colour scheme should apply to illustrations as well as to the coloured elements on the page. Main ELIXIR colours:

- Orange: #f47d21
- Dark blue: #023452

There are also lighter blues and a green (BEWARE: it looks hideous with the orange). Finally there is grey as well. Can post the hex codes for them all if required.

bedroesb:
Are there fonts that we could use? 
First try with the colors:
![image](https://user-images.githubusercontent.com/44875756/86897015-2bae5a80-c107-11ea-986d-93475b1cb924.png)

bedroesb:
The grey is #efefef taken from the elixir website :)

martin-nc:
Hi Bert! I'd use white text in the header because the contrast will be better (accessibility). Looking better, though! We use Lato on the main site. It's a free Google font.

martin-nc:
If you're in the CSS, how about adding slightly more padding to the links in the left menu, to make each item slightly bigger? And maybe a couple pixels more space between the items?

bedroesb:
Hopefully I didn't do to much padding:
![image](https://user-images.githubusercontent.com/44875756/86900937-41724e80-c10c-11ea-8711-58f510c13d62.png)

I also like the title above the left side panel in the elixir-europe website to identify the current section

martin-nc:
Yes, maybe the font size was okay before. The extra padding looks good, I think. It just makes the items easier to tap with a finger, or even find with a mouse cursor. You don't need such fine motor control (important for accessibility, too e.g. people with arthritis in their hands).

bedroesb:
based on #27 I think this can be closed?

Issue #17, opened 2020-07-07
Branding - acknowledgement grants


frederikcoppens:
Which projects / institutes do we need to acknowledge ?
Where and how do we do this ? About page ? Logos on home page ? ELIXIR in template so on each page ?

CaroleGoble:
I added an acknowledgements.md file

bedroesb:
We could do this in the footer if it is not too much

CaroleGoble:
yes  -we need to work out all the funders

bedroesb:
Footer and about page is added during the latest update. Does this give enough possibilities?

bedroesb:
@martin-nc Can you check if we need the noral ELIXIR logo if we already have the converge logo?

Issue #4, opened 2020-07-07
About page


frederikcoppens:
Page to clarify the context in which this was made, acknowledging funding etc.

bedroesb:
Content in the about page is added during the contentathon and visible at https://rdm.elixir-europe.org/about.html

PR #27, opened 2020-07-08
CSS refactoring and changes + about link + footer

bedroesb:
This PR includes:

- Use of Elixir colors #23
- Same Font as Elixir Europe website #26
- No fa-book logo in topnav
- No separate themes css + refactoring to simplify things
- margin/padding in sidepanel
- about button in top-nav #4
- Elixir logo in top-nav
- footer with a Acknowledgements and elixir converge logo #17 

image


martin-nc:
Looks good to me. How about replacing the word 'ELIXIR' in the header and putting the logo in instead? I think it will look nicer. 
white-orange-logo

martin-nc:
Also, how about aligning the 'Copyright' and 'Site last generated' lines to the left and leaving the logo to the right? I don't think that text is particularly valuable to people and the logo makes a nice 'full stop' to the page, with your eye moving from top left to bottom right. We might want to add contact links and legal notices under the footer logo later on as well.

bedroesb:
image

Nice idea!, the Copyright part I was thinking of adding it to a dedicated footer that is spanning the whole width of the page in the elixir lighter blue or grey ?

bedroesb:
Start of a footer: 
image

bedroesb:
Should I go for a full title?
image

martin-nc:
Personally I would say 'yes' but perhaps the word 'research' is important? If that's the case, then I think the title will be too long. We just have to use the acronym in the heading but use the full term on the home page, maybe the first line: "The ELIXIR Research Data Management (RDM) Toolkit ... [say what it contains, how it helps you etc]". Maybe stick with 'RDM Toolkit' until we get more input?

martin-nc:
Note: I would avoid centre-aligning things, like the copyright. To give a clear structure to the page then we need to base everything on a grid, and align left or right to the grid edges. This gives the eye clear edges to follow. I see the logo in the header isn't quite aligned to the left edge of the menu, for example. Perhaps nudge the logo and site title in a few pixels? Also, make sure the items in the footer follow an alignment e.g. the line down the left that hits the left edge of the logo and the left edge of the menu. I'm guessing the template already ensures that. Anyway, just a suggestion!

bedroesb:
@martin-nc I am very grateful for your suggestions ;) I am not really a webdevelopper besides some home projects, so this is really usefull!

bedroesb:
@martin-nc the logo in the header has float: left, so in theory it should snap onto the grid. Since everywhere the same kind of bootstrap container is used in the sections, the grid should be uniform in nav - main part - footer. 
update:
image
image

Since I let the footer appear under the window, I can't take a screenshot of both the navbar and the footer ... :)

bedroesb:
the only thing that is "randomly" aligned now is the example column in the footer

bedroesb:
I took the example away

martin-nc:
@bedroesb Yes, it's taking shape (in my view anyway!). Actually you're probably right about the alignment. Its's just that there's a difference between physical and perceptual alignment. An irregular shape like this logo can be physically aligned but not look quite aligned, so you have to move it. I wouldn't bother in our case, though.

There's quite a bit of space between the top of the footer and the footer text - is that saved for something? Think of the grid again e.g. https://kylebrycegd.files.wordpress.com/2015/02/thinking_with_type_grid_8.gif. You have a constant gutter size between elements in a box and the box edge. So I'd expect the space between the top of the footer and the text in the footer to be similar to the space between the top of the header and 'About', for example. This creates a sense of uniformity and rhythm. CSS frameworks should take care of this mostly, but not in all cases. Maybe try using a constant padding in the heading and footer, or at least the top padding? Oh - I've realised the footer text might be vertically aligned in the middle - how about aligning it to the top instead? Again, this is just my view!

bedroesb:
I have been playing around with it, and the topnav is not an easy one since there are 4 different elements with different hights:
- the logo
- the title
- about etc
- search box

And in the situation I want them all centered, only one of them will have the padding chosen (15px I thought to take) , or I have to align them on for example to the bottom, like in the elixir europe website ( but maybe it will than start to feel like a cheap mockup ;) ):

image

In this situation the search box has the desired padding of 15px

martin-nc:
Sorry Bert - I thought the header was fine. It was the footer text I was suggesting needed repositioning, so that it is nearer the top of the footer. Does this make sense? Do you think it looks better?
rdm-toolkit1

bedroesb:
understood ;) now they are both 15px, the bootstrap standard. also for info boxes etc...

bedroesb:
I should probably adjust the padding in the TOC towards the one in my sidepanel too

bedroesb:
But I think this is already a nice amount of changes for this PR ... :)

martin-nc:
Yes, how about just merging it now? I think it looks way better than the current site, don't you?

bedroesb:
way better!

PR #32, opened 2020-07-10
added initial page on Human Data use case and guidance page on the GDPR

pinarpink:
Created the first pages for GDPR and human data use case. To observe the appropriateness of the amount of content and styling.


bedroesb:
Style tip: # is used for the main title, shich is specified by the title property in the metadata of the markdown. In the content itself, the titles and subtitles are starting from ## to ### and more :)

bedroesb:
I changed it already in the document

PR #36, opened 2020-07-13
Fixed inkscape mixup for links

korbinib:
Fixed inkscape mixup for links


bedroesb:
@korbinib The icons at this moment do not contain links, which makes it sometimes difficult to click on it. Could you make them part of the linked object?

korbinib:
@bedroesb  Will do. Is it more readable with the text outside of the objects now?
 I also would like to accompany the diagram with expandable colour matching boxes with explanations on the side.

martin-nc:
Perhaps use the same font as the rest of the page (Lato)? 

`pointer-events: none;` in the CSS will disable any hover effects on the icons, and will just leave the hover on the whole shape.

PR #38, opened 2020-07-14
Update security_&_privacy.md

robertmand:
Original template, first attempt


bedroesb:
`Is my data sensitive?` is this one of the subtitles of security_&_privacy or is this the main title?

PR #51, opened 2020-07-15
Toollist search/sort/paging

bedroesb:
Added searching, sorting and paging to the tool list:

image


martin-nc:
Are the arrows on the tool names necessary? Maybe it's cleaner without them?

bedroesb:
done + overflow with mobile:

image

bedroesb:
ok that became a huge screenshot... :)

martin-nc:
Looks better to me! 

Would it be useful to make the tags links? Or would that just be too many links on the page?

bedroesb:
I like the idea, but maybe we can open a separate issue about that?

PR #56, opened 2020-07-20
Style changes

bedroesb:
The current H3 color is red and maybe to much taking away the attention from H2 (noted by @frederikcoppens ).

Now:

image

After this PR, H2 main Orange, and H3 weight 700:

image

What do you think @martin-nc

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -13,7 +13,7 @@
 <!-- Custom edits on top of bootstrap -->
 <link rel="stylesheet" href="css/customstyles.css">
 <!-- Navigation title font  -->
-<link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Lato:wght@300;400;700;900&display=swap">
+<link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Lato:wght@300;400;500;600;700;800;900&display=swap">
 <!-- JavaScript  -->
 <script src="js/jquery.min.js"></script>
 <script src="js/anchor.min.js"></script>
--- a/css/customstyles.css
+++ b/css/customstyles.css
@@ -21,9 +21,6 @@ body {
     font-variant-ligatures: none;
 }
 
-.h1, h1 {
-    margin-bottom: 20px;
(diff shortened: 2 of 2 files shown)


martin-nc:
@bedroesb Warm colours like orange and red are tricky to use because they demand attention. Also orange isn't that easy to read against white. How about the default dark grey for h1, a dark blue for the h2s (#055472) and a lighter grey for h3s (#666, perhaps change the '--subtle-text' variable to this)?
Screenshot 2020-07-21 at 11 42 31
Also, I wouldn't use italics in headers.

Still on colours: looks like some of the colour palette is still the default. So the '--accent-colour' variable in CSS is red, which I don't think goes with the other colours. Maybe change that to the dark blue I mentioned above?

bedroesb:
I am having doubts about H4 with: 

```
h2 {
    font-size: 24px;
    line-height: 29px;
    font-weight: 700;
    color: var(--accent-color)
}

h3 {
    font-weight: 700;
    font-size: 140%;
    color: var(--subtle-text);
}

h4 {
    font-weight: 700;
    font-size: 120%;
}
```


image

martin-nc:
Ah, I hadn't noticed the H4 styles. How would it look making the H4 the same grey as H3, but the same size as the normal text? This is the trouble if the H2 is not very big then it's difficult to get a strong sense of hierarchy down to H4. 

Also, how about making the bullet points blue? Here's one approach, but maybe the theme uses another:

```
ul {
  list-style: none;
  margin-left: 2.5em;
}

ul li::before {
  content: "\2022";
  color: var(--accent-color);
  font-size: 1.2em;
  display: inline-block;
 /* Add margins as needed. */
}
```

bedroesb:
image


I have to admit that I find the black H3  one more clear :) :

martin-nc:
Yes, I think the second version works better: blue->black->grey. I think we just have to see what they look like in different contexts. If it was me, I'd just put that live and see what it looks like in different pages, then go through another round of improvements.

PR #67, opened 2020-09-09
Feature update

bedroesb:
- [x] Name of the contributor instead of the id/github username -> consequence is that CONTRIBUTORS.yaml is now in _data folder
- [x] Placing the RDM lifecycle in the sidepanel of the website
- [x] svg rdm cycle imrpovements
- [x] correct top-margin onsublist sidebar

 
image

image

At this moment the rdm life cycle position get's a bit lost after clicking on it on the landing page.
I made multiple variations to indicate the current location in the lifecycle and ended up now giving soder colors to the unselected ones


bedroesb:
@martin-nc and other, I don't know how distracting this can be

korbinib:
Maybe keeping everything in grey with dark grey for the active step could be an option to make it less prominent, if you are concerned about this?

martin-nc:
Yes, I think that's distracting - it's partly the shape, but it's also the colours. Grey is okay but a bit dull. I think blue might look nicer, and beneath the menu to be less distracting:

grey-cycle

blue-cycle

Were we going to colour-code sections, though? Also, I still think it's too big. What happens on mobile? Does it go to the top or drop to the bottom of the main content area?

bedroesb:
image

For the mobile mode I was not sure yet ... definitely needs to be smaller than fullscreen right now

bedroesb:
the colors change while hovering to the once of the original

martin-nc:
Ah - so is the highlight colour the original one as well? I guess that would make sense. Maybe it would be nice to pick up the highlight colour in the main content area, to make the connection clearer:

rdm-colour-code

Some colours (like yellow and the green above) might be tricky. They need to contrast with white, but we could darken them slightly in the text. I haven't done that above. I think a splash of colour would make the site more attractive because remember we aren't going to have many photos! Also, the colour code would only apply on pages related to the RDM. What do you think? If the text colour doesn't work then I think it might be worth having another graphical element like a line or border near the page header that reflects the RDM colour.

bedroesb:
I like the idea, I am just thinking how I could do this elegantly to have page specific styling for 7 pages without to many if else cases :)

floradanna:
I also think the circular shape is distracting in any colours. I would keep the usual rectangular boxes.
If we keep the circle, light grey + dark grey or orange when selected might be nicer, or at least more consistent with the other boxes below (your domain, your problem etc...).

By looking at the website landing page, I have the impression that the RDM life cycle stages are more important than the other sections (your domain, your problem etc...), and I was wondering if this is on purpose or it is only my impression. However, after clicking on one stage of the circle all the sections become visually equally important on the side bar. This could be what seems strange. I don't have a specific solution to propose yet, I will think about it.

bedroesb:
mobile: 
image

bedroesb:
I just thought it was a pitty to have a beautiful diagram on the landing page, and when you click on it you lose your position in it and the whole diagram and the fact that it is a cycle ... :)

bedroesb:
also about the mobile mode: here it now doesn't make sense that your role, your problem etc are above the screen, this should move under the content

bedroesb:
Running example can be found here: https://bedroesb.github.io/rdm-toolkit/planning

martin-nc:
I guess you could try sticking it in the menu like this:

rdm-in-menu

Then it isn't a distraction in other sections that are not based on the RDM life cycle. Your example looks nice, though - perhaps stop there and get feedback on it from more people?

@floradanna I think the idea at the beginning was to have one main way of navigating the site but other ways available, too. This was to stop people getting overwhelmed when visiting the site. In the benchmarking exercise we did in the hackathon, we noticed a lot of sites had home pages packed with links, which was very off-putting! So one lesson from the other sites was "Keep it simple".  But you're right about the RDM losing importance after the home page. Personally, I think it would be nice to keep some continuity and put the RDM diagram in the subpages as well, but in a non-distracting way!

floradanna:
I like a lot this layout. Framing the circle inside the rectangle with the header makes a big difference and it kind of keeps things together. I like that. Like this the circle is nicely integrated with the rest of the menu in the sidebar and it is still clear that this is the main way of navigation.

bedroesb:
stupid thing is that I don't know how to do this in the way the website is now build up. The dropdown lists are based on a yamlfile. If I leave out the subtitles (re-using, plannning, etc) the website can not know that it has to open up the navbar in the RDM section on the planning page + it will be very hacky to integrate at that position the wheel ... It is now possible because the wheel is separate from the navbar

bedroesb:
I will try to think of a way, I start to get some ideas

martin-nc:
@bedroesb  yes, I'm afraid I thought about that but didn't say anything! Could you copy the code for a menu item and use it in a custom block, and then insert this above the menu in the template? So technically it's not part of the menu, but visually it is? You could use the same CSS as in the menu, so the style is consistent. To make it active, you could set the RDM life cycle item to be open depending on the class of the body, or a front matter value. Let's say the body tag has a class of "lifecycle" across all the RDM life cycle pages you could go `.lifecycle .rdm-menu-item ` and then put in styles to show the RDM diagram. Just thoughts, anyway.

bedroesb:
it is alive!

image

bedroesb:
https://bedroesb.github.io/rdm-toolkit/planning

martin-nc:
Well done! See what people think, but a couple of things from me (sorry!): 

- The hover on the life cycle diagram is a bit twitchy. I think you need to turn off pointer events for the icons within the sections. At the moment the hover is activated for each section shape, but not for the icon within the shape. So when the cursor hovers over the shape the background colour changes but when it moves over the icon within the shape the background reverts to the default. Just put `pointer-events: none;` in the CSS for the icons and they won't interfere with the shape hover any more. Sorry if you were going to do this!

- I wonder if it looks better with a border around it. It seems to make it belong to the menu more to me.

bedroesb:
@martin-nc thank you so much for the tip! It annoyed me and grouping it differently didn't help. Fix and try out coming up ;)

bedroesb:
We don't realy make use of border anywhere in the site so I tried it without:

image

bedroesb:
I am not convinced :)

bedroesb:
with the border:

image

martin-nc:
Yes, I think just the border works better. The trouble with circles is they don't form strong alignments with rectangles and squares. There's no straight edge for your eye to follow. The circle without the border felt like it was floating to me, but the border creates an alignment with the rest of the menu and so anchors it there. You could probably just get rid of the top border and join it with the orange 'Data Life Cycle' bit, to make the connection even clearer? Use a negative margin, perhaps. Anyway, just a thought.

bedroesb:
The quickest way I got it working was through an element specific style (to overwrite the styles it was inheriting) :

image

bedroesb:
What I mean is that I don't like this: `<li class="active sidebar_rdm_sub" style="margin: -4px 0px 0px 0px;">`
besides that it does the job: https://bedroesb.github.io/rdm-toolkit/re-using

(might wonna clean your cash if things did not change, I had it :) )

nsoranzo:
Works well and looks very nice to me!

martin-nc:
@bedroesb I wasn't thinking of inline styles. Maybe just a sibling selector: `.sidebarTitle + li`. Anyway, whatever works! New version looks good to me.

bedroesb:
If I don't know what to do I really wonna redo the css of the navigation, both side bar and top nav. But it seems to be working! Thanks for the help.

bedroesb:
i made the class more specific:

```css
ul#mysidebar li ul .sidebar_rdm_sub {
    border-color: #efefef;
    padding: 14px 10px 10px 10px;
    border-radius: 0px 0px 4px 4px;
    border-style: solid;
    margin: -4px 0px 0px 0px;
}
```

PR #68, opened 2020-09-10
 Testing the template

floradanna:
Testing the template proposed by WP1-Best practices group. We would like to check if the template looks nice on the website.


bedroesb:
ready for merge?

floradanna:
I think so. We wanted to see how it looks in the website. If it is not ok we change it.

PR #70, opened 2020-09-16
Updating some pages

bedroesb:
Hello @DanFaria, I saw that you made some templates! thank you for that. At this moment we grouped the pages in separate directories of which every directory contains a template. I moved yours towards there. We already made a lifecycle diagram there, but maybe @floradanna can have a look in the similarities?


bedroesb:
test of contribute section available here: https://bedroesb.github.io/rdm-toolkit/how_to_contribute.html

floradanna:
Could we keep both templates in the repository for now? Just because one template was done by WP1 Best Practices group (@korbinib, @vdda, @rwwh, @nicjar) for DM Expert Network and we wanted to test it on the website, but @DanFaria was not present at that meeting. We haven't had the chance to discuss Daniel's template yet. WP1 Best Practices group meets on 24/09.

bedroesb:
I will name it WP1 instead of Daniel than :)

floradanna:
@bedroesb You can keep the one named "Daniel" as "Daniel" (which was done by the RDM toolkit "content group") and call the other one WP1.

floradanna (approved):

DanFaria:
Hi.
Yes, as Flora said, these templates were done by the RDM toolkit "content group" of editors (myself, Ulrike and Mijke).
We definitely need to discuss things with WP1, but our feel was that some of the content put into the "planning" page (e.g. how to select a DMP tool) should be moved to a page of its own in the "Problem" section.

PR #73, opened 2020-09-17
Emoji support + template fixes

bedroesb:
I changed in the templates the metadata for: `folder: RDM_cycle` because this is a constant and 
`search: false` to disable that the searchbar will look into the template pages

### DIFF
--- a/pages/RDM_cycle/TEMPLATE_RDM_cycle_WP1.md
+++ b/pages/RDM_cycle/TEMPLATE_RDM_cycle_WP1.md
@@ -4,6 +4,7 @@ keywords: [Keyword 1,Keyword 2, Keyword n]
 summary: Fill in your summary or delete this property when not needed
 permalink: TEMPLATE_RDM_cycle.html
 folder: RDM_cycle
+search: false
 contributors: [contributor1, contributor2]
 ---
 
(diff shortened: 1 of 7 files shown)


pages/RDM_cycle/TEMPLATE_RDM_cycle_WP1.md
```diff
@@ -4,6 +4,7 @@ keywords: [Keyword 1,Keyword 2, Keyword n]
 summary: Fill in your summary or delete this property when not needed
 permalink: TEMPLATE_RDM_cycle.html
 folder: RDM_cycle
+search: false
```
floradanna, line 7: what it this "search" ? What's for?

bedroesb: it is to disable that that page will end up in the search results. This line will thus need to be removed when the template is copied ...

bedroesb: If you think it is not a problem the searchbar searches in the Template pages, we can leave the line out off course

floradanna: It makes sense to disable that that page from search, indeed.

Issue #49, opened 2020-07-15
Styling refinements


bedroesb:
- [x] Center data life cycle diagram
- [x] Fix font in data life cycle diagram
- [x] maximum width of tables same as main text
- [x] first column with tool names wider so name fits
- [x] 4 buttons home screen wider 
- [x] remove/remap edit me button in all tools page
- [x] margin problem buttons sidepanel, small shift in sublevel when opening toplevel
- [x] take away link icons in 4 buttons home page
- [x] color of H3 titles
- [x] better spacing between sidenav
- [x] fixing problem with icon in the life cycle diagram (make them non clickable)
- [x] fixing html ugliness in generated pages
- [x] fix white space under footer on index page because of the unfolding flexboxes
- [x] Dots of lists in same blue as H2
- [x] better image margins in content

bedroesb:
Dots in blue suggestion:
```
ul {
  list-style: none;
  margin-left: 2.5em;
}

ul li::before {
  content: "\2022";
  color: var(--accent-color);
  font-size: 1.2em;
  display: inline-block;
 /* Add margins as needed. */
}
```

martin-nc:
Actually we can do this more simply by using the `::marker` pseudo-element:
```
ul li::marker {
  color: var(--blue-xx-dark);
}
```
The browser for `::marker` seems pretty good these days. The lists in the summary boxes seem a bit odd to me: they don't have any space to the left and perhaps the font goes a bit small. Can I suggest a slight refactor?
```
#toc {
  padding: 15px;
  background: var(--light-grey);
  margin: 1.5em 0;
  border-radius: 5px;
  font-size: 14px;
}

/* Indent the lists and reset any other padding */
#toc ul {
  padding: 0 0 0 20px;
}

/* Consistent vertical space between list items */
#toc li, #toc ul ul li {
  margin: 7px 0 0 0;
}

/* Sublists */
#toc ul ul {
  list-style: circle outside;
}

/* Responsive style to replace static width */
@media only screen and (min-width: 990px) {
    #toc {
        max-width: 50%;
    }
}
```
Here is a screenshot:
<img width="687" alt="Screenshot 2020-09-23 at 19 28 58" src="https://user-images.githubusercontent.com/15261745/94055390-c6e4d100-fdd4-11ea-898f-8403c6b9d348.png">

[...]

PR #79, opened 2020-09-22
Improved tool + resource table + CONTRIBUTING + css/icon improvements

### LINKED ISSUE #52: Linking the tags in the toollist to pages
A proposal by @martin-nc in PR #51 , we could make the tags in the toollist clickable. But than we should decide on the tags and to which page they refer to.

### LINKED ISSUE #75: Merge tools and resources tables
@bedroesb The editorial board decided to DO NOT separate tools and resources. We would like to have only ONE table with "tools and resources" together.

bedroesb:
- [x] one excel/yaml list for both tools and resources
- [x] one yaml file listing all allowed tags (for pages, tools and resources,...) 
- [x] linking pages through tags
- [x] coloring the tags (took the same colors as in the RDM cycle wheel
- [x] registry column to link towards bio.tools and FAIRsharing
- [x] updating demo page/documentation in usage
- [x] moving how_to_contribute to CONTRIBUTING in root 
- [x] allowing both fairsharing and Bio.tools
- [x] replace all png icons to svg
- [x] better icons for fairsharing and Bio.tools
- [x] table of content css improvements #49 

image

Preview can be seen here:
https://bedroesb.github.io/rdm-toolkit/all_tools_and_resources.html

close #75 and close #52


martin-nc:
Is it worth swapping the positions of the 'Show [number] entries' field and the Search box? Or even putting them together to the left so you get 'Search' then the 'Show entries' field just after? I just thought the search might be more useful.

bedroesb:
@martin-nc sadly this is not easy to snap due to the used javascript library of bootstrap/jquery. I could change it, but than I lose the possibility to go to newer versions easily.

martin-nc:
@bedroesb Ah, okay. Just on the 'Do we need to make a visual distinction between tools and resources?' question, I think we talked about this briefly last Friday, and decided it's probably difficult to distinguish between a 'tool' and a 'resource' so it's best not to distinguish them. Maybe I've misremembered, though. You could spend quite a bit of time wringing your hands over the definitions of 'tool' and 'resource'!

bedroesb:
@martin-nc thank you for the update! at this moment I was making a distinction on the yaml too, because I need to know whether the registry ID is for bio.tools or for FAIRsharing.... than I should save the full links towards these registries

martin-nc:
@bedroesb I think the idea was to keep the header vague i.e. 'Tools and resources', but I think it's best to check with the others.

nsoranzo:
> I think we talked about this briefly last Friday, and decided it's probably difficult to distinguish between a 'tool' and a 'resource' so it's best not to distinguish them. Maybe I've misremembered, though. You could spend quite a bit of time wringing your hands over the definitions of 'tool' and 'resource'!

Agreed, it can be challenging to distinguish among "tool", "resource", and also "service".

> I need to know whether the registry ID is for bio.tools or for FAIRsharing

In fact, there may be a resource that has both a bio.tools and FAIRsharing ID, and we may want to add more registries in the future (like standards, there are never enough registries :).

bedroesb:
ok maybe I can list the registries as icons in the registry column and allow multiple values in the registry field in the yaml file/excel file

bedroesb:
image

Support for multiple registries and new icons

Issue #19, opened 2020-07-07
How to contribute


frederikcoppens:
Inspiration 

https://github.com/alan-turing-institute/the-turing-way/blob/master/CONTRIBUTING.md
https://fairplus.github.io/the-fair-cookbook/Contributing.html

From discussions:
Three contributor pathway levels
Upper - using github directly, using markdown templates, assigning tags, pull requests to notify editors
Middle - using googledoc templates, simple raising issues and notifying editors, simple markdown add to pre-existing pages
Lower - using googledoc templates, emailing editors ro submitting a notification through a form.

nsoranzo:
xref. https://github.com/elixir-europe/rdm-toolkit/issues/3

nsoranzo:
However a contribution reaches GitHub, should we mandate use of Pull Requests and prevent direct pushing to `master`?

bedroesb:
this is covered in the contributing section of the website https://rdm.elixir-europe.org/how_to_contribute.html

PR #80, opened 2020-09-22
flow chart for github way added

### LINKED ISSUE #83: Fix GitHub signup link
The current link used for GitHub sign-up on the[ "Github way" page](https://elixir-europe.github.io/rdm-toolkit/github_way.html) is pointing a place describing various GitHub pricing options. It could be better if we used the link that is simpler to follow for end-users [https://github.com/join](https://github.com/join)

pinarpink:
Added GitHub way diagram, which was originally developed by @floradanna 
This is further work towards #19
Also fixes #83


pages/contribute/github_way.md
```diff
@@ -13,7 +13,14 @@ permalink: github_way.html
 ## Make a GitHub account
 1. Make a [GitHub account](https://docs.github.com/en/github/getting-started-with-github/signing-up-for-a-new-github-account).
 
-## Announce and discuss your proposal through GitHub "issues"
+## Contribution process
+The process of contribution via GitHub is sketched below. 
+
+{% include image.html file="github_way_flow.svg" alt="Process of contributing via GitHub" max-width="1500"  %}
```
bedroesb, line 19:
```suggestion
{% include image.html file="github_way_flow.svg" alt="Process of contributing via GitHub" max-width="845px"  %}
```
If you go over 845pc the image will get out of the body of the page, causing a scroll bar. I saw that the image is quite wide and with small letters which is not ideal in this case.

bedroesb (commented):
image

Beside the design which we can always do later, I like the simplicity/bigger font of floras chart. 

image

This is I think mainly because it did not contain blocks of text. I can always tomorrow look into tooltips that appear while hovering over an svg image, if that is even possible.

I don't know what the other think of it. But for sure I would be careful with big charts that come out of the body of the text :)

pages/contribute/github_way.md
pinarpink: Hi @bedroesb  the boxes in @floradanna 's google doc did not match the text headings. If you could export her slide as sag please go ahead.

bedroesb:
Yours (the new flow diagram) is off course more complete, maybe we ask the opinion of the others ? :)

pinarpink:
> Yours (the new flow diagram) is off course more complete, maybe we ask the opinion of the others ? :)

I'd suggest @floradanna takes a look when she is back. I think we can ask Munazah to hold the testing until bare basics of all guides are there.

pages/contribute/github_way.md
floradanna: Hi, yes indeed the text in the flowchart doesn't correspond to the paragraphs yet. However, we could keep the text in the image very short, so that we can increase the font size and make the text easily readable. The same short sentence could be the title of a paragraphs, but then we should add a brief text between the title and the bullet points to summarise what that step is about, I think. Anyway, as Pinar said, all the info for testing are there and we can do the rest later.

PR #104, opened 2020-10-08
Image out of content fix + how to add a tool guide + no permalink as metadata + new issue template

bedroesb:
- [x] Adding how to add a tool guide (closes #96) 
- [x] fixing image going out of content
- [x] change the need of a permalink/folder metadata field
- [x] contributors based on names instead of github ID's
- [x] new issue template to fix existing page (closes #85)
- [x] changing Demo page to Markdown cheat sheet as suggested by #112

PR #114, opened 2020-10-12
Home page spacing tweaks

martin-nc:
1. Added spacing around the notice box at the top.
2. Removed top margin of the footer (too much space because the RDM diagram already has a large bottom margin set).
3. Created a variable in the home page front matter called `style`, set to 'home'. I used this to target the home page footer only:
`<footer id="footer" class="site-footer {% if page.style %}{{ page.style }} {% endif %}">`

CSS:
```
.home.site-footer {
  margin-top: 0;
}
```
But it might be handy to give custom styles to other elements on the home page. @bedroesb  is that okay? Maybe there was a simpler way.

Preview of changes here: https://martin-nc.github.io/martinc-rdm-toolkit/


martin-nc:
Pushed from the wrong repo and has unwanted commits in.

bedroesb:
@martin-nc If you want to achieve the same without additional frontmatter, you could change your Jekyll to  <footer id="footer" class="site-footer {% if page.url == index.html %}{{ home }} {% endif %}">.

martin-nc:
@bedroesb looks like a bit of your comment got removed. What was the Jekyll? I'll create a new pull request with the new code in.

bedroesb:
sorry about that, fixed!

martin-nc:
@bedroesb That's interesting - it doesn't work for me. I tried all the options in [this thread](https://stackoverflow.com/questions/28932707/check-if-current-page-is-homepage-on-jekyll) and they didn't work, either. Only the front matter approach worked.

bedroesb:
I am sorry, it is not url, but permalink, I use it for the coloring of the lifecycle wheel: 
```
{%- capture re-using -%}{%- if page.permalink == "re-using.html" -%}st0{%- else -%}st10{%- endif -%}{%- endcapture -%}
```

bedroesb:
just figured out that this logic is deprecated in the latest version of the website due to my permalink changes in #104 .

martin-nc:
Maybe having a `style` variable in the front matter isn't a bad idea. I think the original idea above is poor because it just checks for the presence of the variable. It could check for the value instead `{% if page.style == 'home' %}{{page.style}}{% endif %} `. That means we have an option of styling different elements of the site differently according the page you're on. It does it in a way that's simpler than checking the permalink because the snippet doesn't need to change each page (because you have to put different permalinks in). Anyway, just a thought - not urgent or perhaps necessary for now.

bedroesb:
If we want to do that in the future I agree having a style variable in the attributes is the easiest way to do so.

Small thing i just noticed we might wonna apply your margin changes to all message boxes:

Screenshot from 2020-10-13 15-07-37

PR #119, opened 2020-10-13
Renaming the folder structure

bedroesb:
aligning the directories in pages with what is seen on the website:
- your role
- your domain
- your problem


pinarpink (approved):
Thanks @bedroesb

PR #133, opened 2020-10-15
updating the githubway

bedroesb:
I added many images and removed the preview here links. I also added the linking to issues.


pinarpink (approved):
I prewieved locally looks great @bedroesb

PR #159, opened 2020-10-15
Added Content to Storage section in the Problem page

smza:
Addressed the Publishing data to a repository question?


elinkronander:
This is still a draft

floradanna:
- If this content is still a draft, it would be better to convert it into a DRAFT pull request.
- Please, check the template page and adopt the same style for titles and subtitle (## or ###, no ####).
- Make sure to add contributors names in metadata.

PR #186, opened 2020-10-16
Update collecting.md

siirif:
I did not change anything, but I just want to make AN IMPORTANT NOTE: I think one should really talk here about data collection, not reusing data as we have an own section in the life cycle to reusing data.


siirif:
I have no idea who wrote this page, but it definitely needs someone to write it :)

siirif (commented):
Did our changes go through? We rewrote the whole page..

PR #234, opened 2020-11-23
Rename and hide pages

floradanna:
- Rename "Public Repositories" to "Data Sharing".
- Hide page "Other" in " Your Problem".


bedroesb:
We can merge this when #233 is approved/merged :)

PR #246, opened 2020-11-30
Create data_licence.md

floradanna:
Page about data licence.


floradanna:
Data licence page should be shown under "Your Problem" area.

Issue #257, opened 2020-12-09
Update of Data Life Cycle - Preserve


YvonneKallberg:
**Link to the page of interest**
https://rdm.elixir-europe.org/preserving

**Describe what you would change to the page**
Update of the text, typo fixes

PR #258, opened 2020-12-09
Update preserving.md

YvonneKallberg:
Typo fixes, add a bit of clarifying text why important, unifying the language under considerations
#257


siirif:
We were wondering with my colleague Minna Ahokas about the tool list on this preservation page, that it might not be optimal with the text. And also if there would be some real tools for preservation and archiving, it should not overlap too much with the sharing part (if e.g. elixir core data resources / databases should be listed in the sharing page..). I don't have a solution here, just a wondering if this is something we should fix / think about the tools and tags..

bedroesb:
@siirif tools and resources are not listed anymore on the data lifecycle stages but will be listed in the problem pages these lifecycle stages point to

Issue #285, opened 2020-12-17
Update Data Life Cycle - Collecting


YvonneKallberg:
**Link to the page of interest**
https://rdm.elixir-europe.org/collecting

**Describe what you would change to the page**
Since all other pages in the life cycle has a section 'Why is XX important?' I suggest that also collecting should have one.

PR #286, opened 2020-12-17
Add a 'Why is Data Collection important?' section

YvonneKallberg:
Issue #285
Suggestion of text to the mentioned section, as well as removing a dot in the list of considerations and adding a hyphen to Re-use (since this is how it is spelled in the side menu of the toolkit).


floradanna (approved):
Thank you for adding the missing paragraph.

smza (commented):
Thanks Yvonne. We just need to change the Re-use back  to Reuse, as this is what we decided in the recent meeting.

bedroesb:
good catch @smza , do i quickly fix it?

smza:
> good catch @smza , do i quickly fix it?

yes please! Thanks Bert!

Issue #256, opened 2020-12-09
Main RDM lifecycle image shrinking


CaroleGoble:
**Describe the bug**
Using Firefox, when I do full screen the RDM lifecycle picture shrinks and never recovers unless I refresh the window

**To Reproduce**
1. Browser Firefox
2. Select "full Screen" icon
3. image shrinks

**Expected behavior**
Image doesn't shrink

**Screenshots**
![image](https://user-images.githubusercontent.com/7300655/101611522-85da8100-3a01-11eb-94b7-35d6ac6e8e52.png)

**Desktop (please complete the following information):**
 - OS: [windows]
 - Browser: firebox
 - Version [dunno]

bedroesb:
@CaroleGoble thank you for notifying! @martin-nc is working on a fix :)

PR #296, opened 2020-12-22
Tweak to make the toolkit diagram more responsive

martin-nc:
Addressing the issue that the RDM Life Cycle diagram doesn't resize properly (see #256), at least in Firefox. The preview is here:
https://martin-nc.github.io/martinc-rdm-toolkit/

Try making the browser window small then big again in Firefox. You might need to clear your cache or look in a new private window. The diagram should resize accordingly. It should not shrink and then stay shrunken.

@bedroesb Bootstrap 3 already has some CSS to handle responsive embedded objects like `<iframe>` and `<object>`. I didn't realise! The pattern is:
```
<div class="embed-responsive embed-responsive-16by9 main_rdm">
    <object class="embed-responsive-item" type="image/svg+xml" data="images/RDM_circle_final4.svg"></object>
</div>
```
Notice the `embed-responsive` classes and the `<div>` providing a wrapper.


bedroesb:
@martin-nc That is great! Thanks for teaching me on this ;) Looks good to me! 

(don't forget to pull the changes from master from time to time, I see for example that you still have re-use in the wheel, which became reuse :) )

bedroesb (approved):

PR #301, opened 2021-01-05
Making tools assembly available in sidebar and landing page

bedroesb:
Screenshot from 2021-01-05 15-06-19


bedroesb:
@floradanna Forgot to update the screenshot!

PR #302, opened 2021-01-06
fixed typo in git commands; added additional information for working …

bianchini88:
…with jekyll. Note: the procedure is not compatible with the new version of ruby (3.0.0)

### DIFF
--- a/pages/contribute/working_with_git.md
+++ b/pages/contribute/working_with_git.md
@@ -26,8 +26,7 @@ NOTE: if you already did these steps in the past, start from the `git fetch upst
     ```
 - Create a new branch named after your feature/edit.
     ```
-    git branch -b 'FEATURE_NAME'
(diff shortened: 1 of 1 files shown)


bedroesb:
@bianchini88 thanks for the update! (and good to know about the new ruby)

bedroesb (approved):

bianchini88:
@bedroesb running jekyll fails with a -2 arguments instead of 1-type error. Something similar to this:
https://stackoverflow.com/questions/60198057/jekyll-wrong-number-of-arguments-given-2-expected-1-argumenterror
but the solution proposed did not work. I am working on Ubuntu 20.04, btw. The only way I found to make it work was 
to revert to ruby 2.7.2 (although I am not an expert). I did include that warning so that new contributors such as myself would not waste time on that.

PR #303, opened 2021-01-08
Style guide update + better contributor carousel

bedroesb:
Some additional guides


floradanna (approved):

martin-nc:
Just a couple of comments: I'd say 'full stop' rather than 'point'. Point is for numbers, not text (e.g. 2.1 is 'two point one'). Personally I'd always make file names lowercase, even if there is an acronym. It's easier to remember and looks consistent. That's just my view!

bedroesb:
@martin-nc following the filenames of the pages, I guess you would also lowercase the acronyms in the tags? `it support` instead of `IT support`

martin-nc:
I'd see them as text content of the site and I'd keep the uppercase on acronyms, just like I'd keep spaces between the words in tags. I think they're different to file names. Also, this would make it easier if you ever wanted to auto-generate tag pages, with the tag as title.

floradanna:
We decided to write acronyms with capital letters during an editorial meeting. I would not change all acronyms back to small letters.

floradanna:
I would like to define the way "sub-problem" titles/question should be formulated in "your problem" pages. Which way is best?
* Where can **I** find...? OR Where can **YOU** find...? Where xxxx can **be found**?
* Is **MY** data sensitive? Are/Is **YOUR** data sensitive?
* ...

martin-nc:
I would go with the 'you' option. I see there are quite a few types of subheadings, though. Trying to use 'you' every time might sound contrived. For example, you could change the heading 'What is the best way to name a file?' to 'What is the best way you can name a file?' or 'How do you name a file?', but maybe it would be simpler to say 'How to name a file'. In your example 'Where can I find X?',  I'd just say 'Where to find X' (no question mark - it's not a question). Or are we supposed to use questions?

I know we encourage the use of 'you' in the style guide, but that's really in the text, to encourage people to write the way they  speak. So instead of writing 'Metadata standards are to be found...', which you'd never actually say to someone, you could write 'You can find metadata standards...'. That's more readable and friendly.

floradanna:
Thanks Martin. I think it is a good suggestion.
For Life Cycle pages and Your Problem pages, the templates suggested to use QUESTION form when possible. As result, we had a miscellaneous of "question" forms (you, I, passive form, with or without "?" etc...). Therefore, I would like to give clear guidelines about this. 
The advantage of the "question" form it is maybe that force contributors to focus on a more specific and very practical problem. I think the style of "Frequently Asked Questions" was liked in the WP3, at the beginning at least.
Maybe it is worth it to revise this issue tomorrow.

PR #335, opened 2021-01-18

YvonneKallberg:
Minor typo fix and adding links to more of the Your Problem pages. However, since there's also been discussions about minimizing manual linking, this might not be the way to go? Perhaps all links should instead be removed?


floradanna:
Thanks Yvonne. 

Yes, indeed we decided that it is better to avoid manual linking of internal pages. Therefore, I would not list all pages and topics anymore. Maybe just writing something generic to say that most info needed for writing a DMP can be found in this website.

For the listed tools under "solution", ideally we should add more info for each tool. For instance, this is used in France; this is accessible from any university; this is machine actionable ... or something like that, and then refer to the tools table to see more tools.

Issue #317, opened 2021-01-15
Update DataLife Cycle – Reusing


jmenglund:
**Link to the page of interest**
https://rdm.elixir-europe.org/reusing

**Describe what you would change to the page**

I would like to make the text under "What should be considered for Data Reuse?" more readable, perhaps by splitting the paragraph and adding bullets. I also would like some smaller changes elsewhere and add some new content as well (for example an additional bullet under "Why data Reuse is important?".

PR #336, opened 2021-01-18
Update DataLife Cycle – Reusing

jmenglund:
Updated text on the page about Data Reuse, issue #317.

I have made the following changes:

* Changed heading "Why Data Reuse is important?" to "Why is Data Reuse important?
* New description of what data reuse is
* Added one item to why reusing is important
* Formatted considerations as bullets and reworked the text. Also added two new bullets about versioned data and citing.
* Moved section about training to the end
* Added keywords
* Added contributor


YvonneKallberg:
As far as I can see, this PR fix all gaps that were identified at the January 15 Contentathon.

pages/rdm_cycle/reusing.md
```diff
@@ -1,37 +1,45 @@
 ---
 title: Reusing
-keywords:
-contributors: [Korbinian Bösl, Daniel Faria]
+keywords: [Data reuse, Reusable data, Licencing, Licence, Data integration, Versioning, Citing data, Data citation, FAIR]
```
floradanna, line 3: I suggest to keep only "reuse data" and "licence" and "data integration" and "citing data". Redundant tags are not needed because suggested pages appear in the search box meanwhile the user is still typing. The search box already looks for part of the word, so perfect match is not needed.

FAIR and Versioning seems a bit too far for this page, in my opinion.

jmenglund: Sounds reasonable to remove the redundant keywords! I didn't really know what to put in, since so many pages were still lacking keywords.

jmenglund: I also think it is fine to remove "FAIR" and "Versioning"

jmenglund: I have updated the keywords according to your suggestion.

floradanna (approved):

PR #373, opened 2021-01-22
Your role - Data Steward (3 pages) updated. Your role - Researcher polished

bianchini88:


bedroesb:
@bianchini88 I would make use of the tagging system for example: `{% include pagelist.html tag="data manager" %}` of pages to list pages instead of manual linking towards them.

PR #391, opened 2021-02-01
Adding home page blurb + README update

bedroesb:
Screenshot from 2021-02-01 16-44-22

### DIFF
--- a/index.md
+++ b/index.md
@@ -4,11 +4,9 @@ hide_sidebar: true
 toc: false
 ---
 
-{% include note.html content="We're still developing this website. If you want to see what's coming up next, have a look at the pull requests and issues on [GitHub](https://github.com/elixir-europe/rdmkit). Feel free to [contribute](https://rdm.elixir-europe.org/how_to_contribute.html)!" %}
-
 {% include landingpage.html %}
 
 ### Contributors
-This project would not be possible without the many amazing community contributors!
+This project would not be possible without the many amazing community contributors! RDMKit is an open community project, and you are welcome to [join us](how_to_contribute)!
 
 {% include contributor-carousel.html %}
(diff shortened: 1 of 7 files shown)


index.md
```diff
@@ -4,11 +4,9 @@ hide_sidebar: true
 toc: false
 ---
 
-{% include note.html content="We're still developing this website. If you want to see what's coming up next, have a look at the pull requests and issues on [GitHub](https://github.com/elixir-europe/rdmkit). Feel free to [contribute](https://rdm.elixir-europe.org/how_to_contribute.html)!" %}
-
 {% include landingpage.html %}
 
 ### Contributors
-This project would not be possible without the many amazing community contributors!
+This project would not be possible without the many amazing community contributors! RDMKit is an open community project, and you are welcome to [join us](how_to_contribute)!
```
floradanna, line 10: maybe remove the first exclamation mark (or the second)?

martin-nc:
I like the blurb. I just had three minor points:
1. A portal is literally a doorway and doorways don't guide you, so that sounds an odd sentence to me. Also, I think of 'guides' and 'portals' as different things. A portal with just have links with maybe a bit of blurb about each link. A guide will have more articles and advice about how to use resources. 
2. 'from...till': I think it should be 'from...to' here. 'Till' is used in some expressions like 'from dusk till dawn' but most of the time it's 'from...to'.
3. '...help you with any problem': is that over-promising? I know the 'Your problem' menu is quite long but presumably there are problems not covered? Would 'help you with problems' be less risky?

Anyway, maybe this is all just me, so feel free to ignore!

bedroesb:
- [ ] 1

- [ ] 2

- [ ] 3

bedroesb:
I added a checklist to your suggestions @martin-nc

PR #392, opened 2021-02-01
Adding schema.org metadata to the website

### LINKED ISSUE #2: Applying Bioschemas to the site
What are the bioschemas profiles we should apply to the site?

bedroesb:
this will close #2


bedroesb:
@martin-nc good that you notice, this was for now the selected one liner :)

PR #394, opened 2021-02-02
Added quick-fix redirect to a secure url

martin-nc:
A hacky line of Javascript that may fix the problem that when you go to https://rdm.elixir-europe.org you get a security warning ('Site not secure'), and a redirect doesn't happen. A more permanent fix has been requested.


bedroesb (approved):

bedroesb:
@martin-nc I am afraid it didn't help and the problem exists before the html is loaded ... :/

PR #401, opened 2021-02-08
European flag in footer

bedroesb:
Screenshot from 2021-02-09 17-13-07


nsoranzo:
I don't think the black and white logo satisfies the guidelines: https://ec.europa.eu/info/sites/info/files/use-emblem_en.pdf

bedroesb:
@nsoranzo I used the same flag as in the ELIXIR Europe website, but gut to know, will change it

nsoranzo (approved):
Thanks for following up!

PR #419, opened 2021-02-19
Update tool table with new tools

lauportell:


bedroesb:
hey Laura! I was wondering how you did the .csv file manipulations? Because in normal situations excel or other programs place quotes when there are commas used in the string automatically? (since this whas the main problem here)

bedroesb (approved):

PR #420, opened 2021-02-19
Revised linking to tools

bedroesb:
Instead of manual linking to tools in the text, this PR would it make possible to link towards tools in the tooltable. This also allows us to visualize the tool links in a different way than the external links.

Screenshot from 2021-02-19 15-07-18


nsoranzo:
I wouldn't use the red colour for external links, maybe you can try the "little arrow pointing out of a box" icon: https://developer.mozilla.org/en-US/docs/Learn/CSS/Styling_text/Styling_links

martin-nc:
I guess I worry that if you made some links look different to others would that be confusing to a new user? Could they guess what would happen when they click on them? And I also might find it tiresome to search for my place again after being jumped down to the table. If it was me, I'd have no link at all in the text, or just link to the resource website. That feels simpler to me. Presumably they get context about the resource from the text and don't need to see it in the table, and click twice to visit the resource? Plus, if people contribute to the page, will they know how to put anchor links to the table in?

martin-nc:
I was assuming the red links will jump you down to the entry for the resource in the table at the bottom of the page. Maybe I got that wrong!

bedroesb:
@martin-nc you are right, they act like buttons to scroll down. @nsoranzo these would not be external links but would lead you to the corresponding tool in the tooltable. They don't have to be red, but it would be nice to distinguish them in a way. Especially in a way that the user is not surprised that it leads them to the tool in the page and not to another website

bedroesb:
@martin-nc not manually linking in the text to the external source of the tool would be the best than... Because otherwise there is no need for a tool table :) But I wonder myself than, what is better, having a button that moves you down to the source, or you having to manually lookup for the tool in the table/or the internet itselve ...

martin-nc:
@bedroesb Can you remind me what the purpose was of linking to the tools table from the text? I've forgotten. Was it to have links automatically generated as much as possible?

bedroesb:
It was a solution I came up with for 2 situations: 
- Where contributors manually link to the tool resource when they talk about it in the text. 
   - In this case the url can be out of sync with the one in the table
   - People will not feel the need to go the the tooltable because they can just click immediately on the toolname if they are interested
  - Gives the possibility to visually distinguish between tools and external references.This case tools can fell like a separate  identity that is part of the tooltable
-  If contributors talk/mention about a tool and they do not link towards source it
   - In this case the user if interested in the tool will have to scroll down and have to search for the tool in the sometimes very long table, and probably
   - First time visiting users might not know that on every page at the bottom you can find a filtered tooltable

floradanna:
We will implement the technical solution to keep the link to a tool in sync between Text and Tools table, after the first release.
The transition to this new system will be supported by scripts that will identify links for tools present in Text and Tools table or only in one of the two places.

PR #431, opened 2021-02-26
More links for workflows

stain:


bedroesb:
Hello Stian, thank you for your contribution! We are actually having the discussion with the editors about hard linking towards tools. Because if we link in the text towards the external page of a tool:
people will not use the toollist at the bottom of the page + urls might get out of sync. Part of the discussion is taking place #420

stain:
Hi, Bert, I still think it would benefit the text to have regular hyperlinks, specially on foundational things like Docker. Perhaps a way to include sub-sections in different parts of the text would help as well, but it will change the flow of the text.

bedroesb:
I agree @stain , we can merge this PR since this should not be stalled util we make a decision on this. We will have to adapt other pages with either decision we make anyway.

bedroesb:
@floradanna or @DanFaria is this good to go?

floradanna:
Yes, I think we should merge this PR.
@stain A paragraph about Docker (containers in general, actually) is still missing; it is one of our gap. We encourage the creation of an additional sub-problem to this data_analysis page to talk about containers. The sub-problem could be "How to make code more FAIR ?" Or something similar. The sub-problem title should be about "when or why or for what problem" you use containers and the content of the sub-problem can explain it.

If you or other data analysis expert in your group are willing to write such paragraph, it would be very welcome. Please, let us known or open an issue+PR about it.

Thanks!

stain:
Thanks for suggestion, @floradanna - added on containers in #442

PR #442, opened 2021-03-12
About Docker, containers etc

stain:
Perhaps this is becoming large enough to become its own section?

Note that I've used plenty of links as they are in context to stable software, and the flat listing would not be able to say things like "uDocker is for user-space"


floradanna:
@stain Thanks for your contribution. It is very valuable.
I will merge your PR, but I will then restructure the paragraphs according to our template (sub-problem: description, considerations, solutions).
I will also add your name to the contributors list.

If some of the tools and resources you mentioned need to be added to our tool list, please fell free to add them as explained [here](https://rdmkit.elixir-europe.org/tool_resource_update.html).

stain:
Thank you for your restructuring in #445, @floradanna, this looks very nice!

PR #456, opened 2021-03-23
Theme improvements

bedroesb:
With this pull request it will be possible to serve pages on different URLs than `root` allowing:

`https://rdmkit.elixir-europe.org/your_domain/marine_metagenomics.html`

\+ more css variables for easy reuse of the theme ( will be improved if scss is used)

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,5 +1,5 @@
 url: rdmkit.elixir-europe.org
-repository: https://github.com/elixir-europe/rdmkit
+repository: elixir-europe/rdmkit
 
 topnav_title: RDMKit
 # this appears on the top navigation bar next to the home button
--- a/_config.yml
+++ b/_config.yml
@@ -23,9 +23,7 @@ port: 4000
 # the port where the preview is rendered. You can leave this as is unless you have other Jekyll builds using this same port that might cause conflicts. in that case, use another port such as 4006.
 
 exclude:
-  - .idea/
   - .gitignore
-  - vendor
   - CODEOWNERS
   - var/
   - README.md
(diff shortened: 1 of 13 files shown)


martin-nc:
Looks okay to me! Makes sense using the relative urls.

On the CSS variables: are they overriding Bootstrap ones or are they new ones? Might be simpler to override Bootstrap ones, that's all, especially if we move to SASS.

bedroesb:
Do you mean the color variables like `--bs-blue` and so on?

martin-nc:
Sorry - yes I did.

bedroesb:
problem is that if I now overwrite for example `--bs-pink` and I use this theme for another project, I will have to overwrite/use another bs-color variable. Or maybe I see things wrongly :)

martin-nc:
I was just thinking in terms of SASS. I think Bootstrap declares it's colours in a SASS map (like an associative array). Then the variables in this map will be passed through certain functions to generate colour schemes for different elements. If you override these variables then you can still take advantage of the built-in Bootstrap functions for them. If you create your own colours then I'm not sure if you can take advantage of these functions so easily.

If you wanted to change the theme slightly for another site you could choose choose different values for the variable in each site, I guess? I'm assuming we're just talking about colour changes. I think you would have a main SASS file where you would import Bootstrap functions and mixins, then have a code block where you override Bootstrap SASS variables, then you import the other Bootstrap SASS files. So you just have to change that code block (a SASS map) for each site. Anyway, let's go with what there is for now!.

bedroesb:
@martin-nc you seem to describe what I want, lets have a look into this we we move to scss!

PR #457, opened 2021-03-24
Adding extra button on the home page for All tools

martin-nc:
I've added an extra button below the RDM Life Cycle. On wider screens there isn't enough space for "All tools and resources", so I've just put "All tools". Not sure if that's okay?

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -27,7 +27,8 @@ exclude:
   - CODEOWNERS
   - var/
   - README.md
-  
+  - vendor
+
 feedback_subject_line: RDMKit feedback
 
 contact_email: rdm-editors@elixir-europe.org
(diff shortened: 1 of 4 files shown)


bedroesb:
@martin-nc and what if we drop the sidepanel? ;)

martin-nc:
I wondered about the behaviour of the buttons. The drop-downs can be wider than the buttons, which I think is okay because it makes them shorter. But I think it would be clearer if the button had an active colour so it's really clear which dropdown is open. This could be the maroon, which at the moment is a hover state over the dropdown links.
 
Screenshot 2021-03-24 at 10 36 24
People won't see this hover state on mobile. Perhaps it would be better to have a grey as the hover state and maroon for the active button?

martin-nc:
Sidebar removed. "All tools and resources" can fit in now but the text will wrap between about 1100 and 860px screen widths, so not great :o(

Preview: https://martin-nc.github.io/martinc-rdm-toolkit/

Issue #382, opened 2021-01-26
Add "Is access and benefit-sharing (ABS) legislation relevant for my research?" to Your Problem – Compliance Monitoring


jmenglund:
**Link to the page of interest**
https://rdm.elixir-europe.org/compliance_monitoring.html

**Describe what you would change to the page**
Based on discussions in issue #364, I would like to add a problem concerning access and benefit-sharing (ABS).

floradanna:
Thanks for opening this issue.

Would it be possible to highlight in the title what is the problem that ABS could solve, instead of saying that ABS is a solution for some problems?
For instance maybe: "How can you access genetic resources of a country?". Or some similar kind of question.

Please, let me know if my example is not clear enough.

jmenglund:
Yes, that is a good point! What do you think about something like "How can I ethically access genetic resources of another country?". I would prefer to narrow it down to ethical aspects, so that we don't have to write about other problems involved in getting access to resources.

floradanna:
Yes, it sounds good to me. Thank you.

Issue #364, opened 2021-01-19
Gap analysis for remaining pages (post Jan2021 contentathon)


pinarpink:
**Description**
Pages for which gap analysis has not been done yet:

- [ ] Storage
- [ ] Data classification
- [ ] ...

jmenglund:
I just made a PR (#362) with some updates to Data classification. There might still be things left to do, though...

floradanna:
- [ ] Almost all "Your Problem" pages need a revision.
- [ ] Identified missing pages/topics:
* FAIR principles and definition of DATA in this RDMkit
* Archiving, Cold vs Active storage
* Informed consent
* Cloud 
* Collaborative research
* Federated computing
* Data Documentation
* File format open vs proprietary

jmenglund:
I think it would also make sense to write somewhere about the Nagoya protocol and EUs ABS (Access and Benefit Sharing) legislation, https://ec.europa.eu/environment/nature/biodiversity/international/abs/

veraort:
> I think it would also make sense to write somewhere about the Nagoya protocol and EUs ABS (Access and Benefit Sharing) legislation, https://ec.europa.eu/environment/nature/biodiversity/international/abs/

I organized a Nagoya Information meeting in Germany a few years ago. I have some content I could add if we want to build a page on this topic.

jmenglund:
That's great @veraort that you want to contribute on this in my mind very important issue! I'm also happy to contribute some content if people think it a good idea to include the information on the site.

After giving it some thoughts during the weekend, my suggestions would for the moment be to add a separate problem page called "Access and benefit-sharing (ABS)" with a problem called something like "Is access and benefit-sharing (ABS) legislation relevant for my research?"

What do you think @floradanna @veraort?

veraort:
> That's great @veraort that you want to contribute on this in my mind very important issue! I'm also happy to contribute some content if people think it a good idea to include the information on the site.
> 
> After giving it some thoughts during the weekend, my suggestions would for the moment be to add a separate problem page called "Access and benefit-sharing (ABS)" with a problem called something like "Is access and benefit-sharing (ABS) legislation relevant for my research?"
> 
> What do you think @floradanna @veraort?

I like the idea and the working title

floradanna:
It sounds interesting. We have to see what is the best place for this topic in the RDMKit, indeed.
I am not familiar with Nagoya protocol and ABS, is it something valid for Biological Diversity only or it applies to every domain?
[...]

PR #458, opened 2021-03-25
Update compliance_monitoring.md

veraort:
I made a new section on the compliance monitoring and measurement page with the information I had collected from our seminar and what Markus Englund had already compiled in the issue #364 and #382. I have therefore simply listed Markus Englund and myself as contributors. I hope you agree with this @jmenglund.

### DIFF
--- a/pages/your_problem/compliance_monitoring.md
+++ b/pages/your_problem/compliance_monitoring.md
@@ -1,7 +1,8 @@
 ---
 title: Compliance monitoring & measurement
 tags: [plan, researcher, data manager, policy officer] 
-contributors: [Christophe Trefois, Wei Gu, Pinar Alper]
+contributors: [Christophe Trefois, Wei Gu, Pinar Alper, 
+Markus Englund, Vera Ortseifen]
 description: measure compliance to data management regulations and standards.
 ---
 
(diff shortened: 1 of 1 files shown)


floradanna:
@veraort Thank you very much for this important content.
I am going to re-format a bit the layout and the style of the text, according to RDMkit way of writing.

I would like to add a practical example about the problem "How can I ethically access genetic resources of another country?". So, could you tell me if the following example would be appropriate or propose some other examples please?

Example: A researcher wants to re-use genetic data from an organism that is present only in a foreign country (animal or plant, etc) for academic research purpose (the research might generate an opportunity for intellectual property).

veraort:
Dear @floradanna 
Thanks for the reply. It's nice that the content helps. Add a practical example is a great idea and it is already a suitable example. I would rephrase it a bit or add a second example, because it does not only concern the "re-use" case and Compliance measurements are added already under the first part "Planning - Problems to be addressed at this stage".

A researcher wants to re-use _or access new_ genetic data from an organism that is present only in a foreign country (animal or plant, etc) for academic research purpose (the research might generate an opportunity for intellectual property).

or

"A researcher from a country, which is party to the Nagoya Protocol, wants to access or re-use genetic resources and the associated data for research and use after 12 October 2014  from a foreign provider country. "

But maybe you can think of something better. But I would have both cases in there - re-use and new access.

Maybe we could also add this publication: "Explanation of the Nagoya Protocol on Access and Benefit Sharing and its implication for microbiology PMID: 28086069 DOI: 10.1099/mic.0.000425 " because they add "Generic Guidance: What biologists must do"

floradanna:
Great, thank you for the suggestions!

PR #466, opened 2021-03-25
Front page changes

bedroesb:
This PR brings a new frontpage layout:

Screenshot from 2021-03-25 23-14-02

as you can see this does not really fill the space in a nice way, so I changed the script into:
Screenshot from 2021-03-26 12-54-20

preview: https://bedroesb.github.io/rdmkit/index.html

@DanFaria was willing to write some descriptions about the sections so visitors understand why we chose to have these sections.


martin-nc:
Second version looks nice to me. I don't think the version with the life cycle in the box next to 'Your role' really works. If we are going for the 'every section is equal' approach then it probably needs more thought, and more time than we have now! That's just my view.

PR #472, opened 2021-03-26
added tag and fixed section title

pinarpink:
@bedroesb this page is ready to be added to the side bar with the label "COVID-19 Data Portal".


bedroesb:
@pinarpink I guess I can also add the tag covid-19?

pinarpink:
@bedroesb ok sure let's have a dedicated **covid-19** tag

PR #476, opened 2021-03-26
Update compliance_monitoring.md

floradanna:
Formatting, styling and improving Nagoya Protocol section

### DIFF
--- a/pages/your_problem/compliance_monitoring.md
+++ b/pages/your_problem/compliance_monitoring.md
@@ -1,7 +1,7 @@
 ---
 title: Compliance monitoring & measurement
 tags: [plan, researcher, data manager, policy officer]
-keywords: [metrics, legislation]
+keywords: [FAIR metrics, legislation]
 contributors: [Christophe Trefois, Wei Gu, Pinar Alper, Markus Englund, Vera Ortseifen]
 description: measure compliance to data management regulations and standards.
 ---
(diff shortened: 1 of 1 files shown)


floradanna:
@veraort Hi vera, I merged your PR and then I opened this PR to make some changes.
I re-structured your text, fixed few links and added some additional information. Please, check that it is still ok for you and let me know.

I also wanted to ask you how the Nagoya Protocol deals with digital data (NGS of genomes etc..)? I found this: https://www.cbd.int/dsi-gr/ , but it seems like work in progress.
Also, if there are downloadable templates for PIC and MAT, please feel free to provide these resources.
Thanks!

veraort:
Hi @floradanna - Thank you very much for your efforts. Regarding PIC and MAT-templates: According to my information, there is one ministry per country that offers approaches via e.g. online forms. But I'll do more research on that. 
For digital sequence informations - It was noticed, that some Parties of the Nagoya Protocol have adopted domestic measures that regulate the access to and use of digital sequence information on genetic resources as part of their access and benefit-sharing frameworks, but there is big discrepancy between the different parties. Therefore, an ad hoc technical group was set up to work out general guidelines. As far as I know, no guidelines have been adopted so far. That was already at the end of 2018 / beginning of 2019 - so I would expect updates to come this year. If something happens, I will put the information on the page. For more: https://www.cbd.int/doc/decisions/cop-14/cop-14-dec-20-en.pdf

floradanna:
Thanks @veraort . So, I will merge this content now and you can add more info in the future, if you want.

PR #477, opened 2021-03-27
Collapse sidebar in mobile mode

bedroesb:
Finally it is here: no giant navigation panel anymore in mobile or smaller windows!
With bootstrap it was also possible to do this with almost no code.

The current version can be previewed [here](https://bedroesb.github.io/rdmkit/index.html)

~At this moment the button has the same style as the other hamburger menu but on the left side + sidebar~

Screenshot from 2021-03-27 18-23-15
Screenshot from 2021-03-27 18-23-26
Screenshot from 2021-03-27 18-23-38


martin-nc:
I was thinking of doing this, too! It would be nice to have before release. It looks good to me, except I wouldn't call it 'Sidebar'. Knowing where something is doesn't tell you what's in it! If someone doesn't know the site it could be for some peripheral content. Really, it's the main menu, but we've already got a menu where the main menu usually is, at the top! So you could call it 'Main menu' to make it obvious this is the one to click, not the top one. 'Menu' would be shorter but a bit confusing when there's another one.

'Browse resources', 'Find resources' or 'Browse the site' are possibilities, but for me 'Main menu' was best.

On the button: I wonder if having the button to the right instead of left would be less distracting if you just wanted to read the content? And would it give a cleaner separation from the content if it had an outline like a button, even when it's not active? The header button doesn't need an outline because the header separates it from the content, but when the menu is in the content area perhaps it's looks a little unstructured and floating? If the button was aligned right I'd put the hamburger menu to the right of the text.

bedroesb:
@martin-nc 

Screenshot from 2021-03-27 20-45-37
Screenshot from 2021-03-27 20-46-09

Screenshot from 2021-03-27 20-50-40


the black border only appears when I take the screenshot. 
I like to be honest if it is on the left side, to distinguish it more from the top navigation hamburger menu + the two hamburger menus don't align perfectly which triggers me , just my opinion!

bedroesb:
Screenshot from 2021-03-27 21-12-42

With the icon being the same size it bothers me less ! :)

bedroesb:
This is also possible:

Screenshot from 2021-03-27 23-01-03
Screenshot from 2021-03-27 23-01-11

bedroesb:
Screenshot from 2021-03-27 23-08-18

bedroesb:
I guess an different 'active' color would be better to distinguish from the other active ones

martin-nc:
I like it with the 'Main menu' text - that's clearer for me. I see what you mean about the right alignment, though. Maybe it wasn't such a good idea. I think it's better on the left. The idea I had originally was like your last idea, but wrapping the whole menu in a div with a background (dark grey?) or a border, so that extent of the menu is obvious when it's open, and the menu is joined to the button like an accordion. Maybe that's not necessary. It's hard to tell without using it.

Both the button and full-width options take up quite a bit of space. I wonder if the button (full width or not) could be moved nearer the top bar and away from the text, so it is more associated with site navigation elements? The space between the top bar and text could be the same as now, but the button is one third down the space rather than half. Does that make sense? Will that work better?

I might slightly prefer the full width version - I guess it's easier to tap! And yes, a different highlight colour would be good. Would the dark blue be too strong?

bedroesb:
Screenshot from 2021-03-28 23-50-07

If we go for a dark background, quite some css will have to be added to let it look decent. A border is I think not an option in combination with the border around the wheel.

This is with the subpages having a white background:
Screenshot from 2021-03-28 23-51-46

This is a quick example with the sublists being white:
Screenshot from 2021-03-28 23-55-38

I agree on the spacing and will look into that tomorrow. Maybe after we merge #466 I can preview it on my fork to test out things

bedroesb:
I start to like this:

Screenshot from 2021-03-29 00-22-23
Screenshot from 2021-03-29 00-22-30

I have just a bit of a margin problem, I need margin under the button, but not when it is unfolded. But I can't add margins or paddings to the collapsing div because that acts weird. And I don't manage to overwrite the bootstrap margin (mb-4) in case the collapse class is added to the button.

martin-nc:
Could you do something like `.button.active + .menu` and remove the top margin to the menu when the button in an active state? Or give it a negative top margin if it already has no top margin? I'm assuming the button and the menu are siblings.

martin-nc:
By the dark background I meant something like this, with possibly a drop-shadow. But let's not worry about it for now. What you have above is good, too, especially if the spacing can be fixed.

mobile-menu2

PR #480, opened 2021-03-29
Adding logo's to style guide

bedroesb:
These logo's can be used for presentations and so on

Screenshot from 2021-03-29 21-53-15


martin-nc:
I don't think the logos belong to the style guide, do they? The style guide is for people writing for the site. The logo is more for people creating things like publicity or slides about the RDMkit. It's probably needs its own page, like 'Media' in the 'About section'. This could also be the place for other publicity about the RDMkit (leaflets, posters etc), if such things were created.

bedroesb:
@martin-nc good point! I will maybe make a new sidepanel for the about? With about, Media, editors?, to contributors (future), ...

martin-nc:
@bedroesb Ah, I thought the About page had the same side panel as 'Contribute'. I misremembered. Maybe it should? Is that what you mean? Add 'Media' to the Contribute menu and use the Contribute menu in the About page?

bedroesb:
I had a new sidebar in mind like:

Screenshot from 2021-03-30 17-20-07

with organisation (or another title) having following subpages:

Screenshot from 2021-03-30 17-22-08

PR #481, opened 2021-03-30
Update landingpage.html

DanFaria:


martin-nc (approved):
I just wondered if the last sentence was necessary? Not a big deal, just a comment.

DanFaria:
I felt it was relevant to provide a brief explanation of the navigation paradigm, and Bert agreed with it, but if nobody else feels it is needed, we can just scratch this pull request.

bedroesb:
I think it serves nice transition towards the structure of the side. But that is just my opinion :)

martin-nc:
Let's leave it in then.

PR #482, opened 2021-03-30
Revised section descriptions

DanFaria:


bedroesb:
The descriptions are demonstarted here: https://bedroesb.github.io/rdmkit/index.html

PR #494, opened 2021-03-31
Add the Common Workflow Language standards

mr-c:


floradanna:
Hi Michael, thank you for your contribution.

Since CWL can be used in every domain (it is not domain specific), we shouldn't tag it with domain tags but only with 
"IT support, researcher, data analysis". 

You can edit your PR with the suggested tags or one of the editor can also do it, if it ok for you.

mr-c:
> Since CWL can be used in every domain (it is not domain specific), we shouldn't tag it with domain tags but only with
> "IT support, researcher, data analysis".

Okay. I didn't know if not having a domain tag would result in less surfacing of the CWL standards in the toolkit. I'll fix it here, thanks!

bedroesb (approved):

PR #505, opened 2021-04-08
text about tiered storage.md

smza:
@floradanna please have a look. I have added some text about the tiered storage.


floradanna:
It looks great to me. Thanks!

I would only suggest to put a link to a page that extensively describe the tier classification (wikipedia or similar) when you mention the word "tiered storage" the first time, if such page exist.
I would also explain what the acronymous SSD stays for.

I think you can merge this PR. Thank you.

floradanna (approved):

PR #549, opened 2021-05-14
Improving accessibility

### LINKED ISSUE #484: Suggested accessibility improvements
I'm just writing an accessibility statement for the site and have done a small accessibility audit. There are a few issues that probably need fixing:

- [ ] **Skip links:** if someone is navigating the site with the tab key we need hidden links that will allow them to skip past lists of links, which are tedious to go through with a tab key. So we need hidden links to 'Skip to the content' (and miss the main menu), 'Skip the tools table' (otherwise you have to go through every single link with your tab key!), 'Skip the contributors' (again, loads of links). Skip links should be visible when they become focused using the tab key.
- [x] **Heading levels:**  the headings in the footer are `<h6>`. You can't skip heading levels in HTML. After `<h2>`, for example, you have to have `<h3>`, no matter what it looks like. Some assistive devices rely on the heading hierarchy to navigate. So we need the change the footer headings to `<h2>` and use CSS to make them look smaller. Note: you can skip levels upwards e.g. `<h6>` to `<h2>`, but not the other way round.
- [x] **The button for Training materials** fails the contrast checker.
- [ ] **The links in the life cycle wheel** in the menu and home page don't focus when using tab navigation. You don't know where you are when navigating the wheel with a keyboard. 

For reference:
 - [WAVE accessibility checker ](https://wave.webaim.org/)
 - [UK government accessibility checklist ](https://www.gov.uk/government/publications/doing-a-basic-accessibility-check-if-you-cant-do-a-detailed-one/doing-a-basic-accessibility-check-if-you-cant-do-a-detailed-one)
 - [W3C accessibility checklist](https://www.w3.org/WAI/test-evaluate/preliminary/)

And also as a reminder, we are required to meet the 'AA' level of the [The Web Content Accessibility Guidelines 2.1](https://www.w3.org/TR/WCAG21/).

bedroesb:
- adding skip to content link
- adding skip to footer link
- adding skip tool table
- more alt texts in images
- fixed no robots/sitemap xml when page has `search_exclude = true`
- better tab navigation of the data life cycle wheel
- h1 title on homepage
- nav sidebar + better search
- better main
- hover grey more contrast
- fixing relative links
- usefull link texts
- unique ids in both links and paths of svgs

Preview can be tested [here](https://bedroesb.github.io/rdmkit/)

closes #484


martin-nc:
Looks great! I guess this means we need to update the [accessibility statement](https://rdmkit.elixir-europe.org/accessibility.html) and resolve [#484](https://github.com/elixir-europe/rdmkit/issues/484) as well? @bedroesb do you want me to update the accessibility statement or are you okay doing that?

bedroesb:
@martin-nc It would be great if you could do that, if you have time off course!

PR #555, opened 2021-05-18
link fix + edit me fix

bedroesb:


bedroesb:
@korbinib I noticed that if you do not add www. it will link it to an internal page, which in this case does not exist.

korbinib (approved):
Thanks for spotting and fixing this.

bedroesb:
Also thanks to @floradanna for spotting this!

PR #559, opened 2021-05-19
Minimal tool assembly template

korbinib:
This is a minimal stub template for the upcoming contentathon. Maybe we can refine it after the event.


floradanna (approved):

smza:
@korbinib looks good to me, I would just suggest to keep a link to the existing tool assemble like the NeLs one, to help the writers. Just a simple sentence  for  e.g. see "NeLs page"  for reference.

smza (approved):

korbinib:
Added a reference for an infrastructural example (NeLS)

PR #573, opened 2021-05-27
Update nels_assembly.md

bianchini88:
Added some more info on DSW. Polished text.


korbinib (approved):
Replaced TSD with NeLS in the DSW section.

PR #586, opened 2021-06-01
Start of tools assembly index page

bedroesb:
A try out in automatically listing the assemblies on an index page making use of the metadata in the pages:
- title
- tags
- tools used in page
- summary

Screenshot from 2021-06-01 17-05-55


korbinib:
I think this could work (provided we have sufficient metadata) as long as there are not too many assemblies. It think we need a round of tagging and tooling (including the NeLS assembly).

I am still a little sceptical how it would scale for more assemblies. Would a filter based on a tag cloud (or similarly) work?

korbinib (approved):

bedroesb:
I think we could include filters in the future yes! But indeed metadata is important here.

PR #590, opened 2021-06-04

lauportell:
Add the H2020 template for data management plans


floradanna (commented):
I see that the link brings you to the data management page of the EU funding page. It is definitively useful, but user might expect to directly download a template maybe?

We could re-phrase as "From the [H2020 online manual], you can read DMP guidelines and download the H2020 DMP templates in different format". 

Or something similar. What do you think?

lauportell:
> I see that the link brings you to the data management page of the EU funding page. It is definitively useful, but user might expect to directly download a template maybe?
> 
> We could re-phrase as "From the [H2020 online manual], you can read DMP guidelines and download the H2020 DMP templates in different format".
> 
> Or something similar. What do you think?

Yes I think that is a good idea as well. Maybe I can specify that at the end of the page there is the link to download the template

floradanna (approved):

Issue #597, opened 2021-06-11
Adding Person and Organisation metadata to website


bedroesb:
Adding Bioschemas for [Person](https://bioschemas.org/profiles/Person/0.2-DRAFT-2019_07_19/) and [Organisation](https://bioschemas.org/profiles/Organization/0.2-DRAFT-2019_07_19/)

bedroesb:
proposal by @jburel to add schema.org for events

jburel:
* https://github.com/elixir-europe/rdmkit/pull/599#discussion_r652455333
* https://github.com/elixir-europe/rdmkit/pull/599#discussion_r652496430

PR #599, opened 2021-06-11
Events on front page

bedroesb:
This PR will add events to the frontpage. The events will only be visible when there are events upcoming.
Events are made visible using JavaScript so no new build of the website is needed. Info is taken from an events.yml file in the data directory.

Screenshot from 2021-06-11 19-34-42

A preview can be found here: https://bedroesb.github.io/rdmkit/

### DIFF
--- a/_data/events.yml
+++ b/_data/events.yml
@@ -1,7 +1,7 @@
 - name: Contentathon
   startDate: 2021-06-23
-  startTime: 9h00 CET 
+  startTime: '9:00'
   endDate: 2021-06-24
-  endTime: 13h30 CET
-  description: We would like to invite you to highlight your set of data management tools as a tool assembly in the RDMkit and describe how to use it, so others can do the same.
+  endTime: 13:30 CET
+  description: We would like to invite you to highlight your set of data management tools as a tool assembly in the RDMkit and describe how to use it, so others can do the same. (two half days)
   location: Online
--- a/_data/events.yml
+++ b/_data/events.yml
@@ -1,11 +1,7 @@
-- name: Contentathon Day 1
+- name: Contentathon
   startDate: 2021-06-23
   startTime: 9h00 CET 
   endDate: 2021-06-24
   endTime: 13h30 CET
   description: We would like to invite you to highlight your set of data management tools as a tool assembly in the RDMkit and describe how to use it, so others can do the same.
-  location: Zoom
-- name: Contentathon 2
-  startDate: 2021-08-23
-  description: Adding new tool assemblies to the RDMkit, it will be fun!
-  location: Zoom
+  location: Online
--- a/_data/events.yml
+++ b/_data/events.yml
@@ -3,5 +3,5 @@
   startTime: '9:00'
   endDate: 2021-06-24
   endTime: 13:30 CET
-  description: We would like to invite you to highlight your set of data management tools as a tool assembly in the RDMkit and describe how to use it, so others can do the same. (two half days)
+  description: We would like to invite you to highlight your set of data management tools as a tool assembly in the RDMkit and describe how to use it, so others can do the same (two half days).
   location: Online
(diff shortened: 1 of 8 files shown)


_data/events.yml
```diff
@@ -0,0 +1,11 @@
+- name: Contentathon Day 1
```
jburel, line 1: Was the idea of using json formatted according to schema.org explored? The content can then be validated using  https://search.google.com/test/rich-results

bedroesb: I putted recently adding schema.org for the contributors on the agenda, but this could indeed also be used for the events thanks!

jburel: If there is an issue, happy to have a go at it if it helps

jburel: @bedroesb I had a go last night at using schema.org for contributors (see https://github.com/jburel/rdmkit/tree/contributors evaluation at that stage) I was curious about the logic in place to handle contributors in https://github.com/elixir-europe/rdmkit/blob/master/_includes/contributor-page.html and other contributor pages.
The approach taken will no longer work when using schema.org.

Is there a reason not to include all contributors in a json file, this could be a second json if required. This will also avoid propagating the logic to the multiple contributor pages

bedroesb: Hello @jburel, I know it is not the nicest logic, but this Is the way that I found to collect all the contributors from the pages (which do not all exist in the yaml file) + al the contributors that exist in the yaml file (which are not all mentioned in the pages). The yaml is only used to extra info or to add contributors that do not contribute to a specific page. This is proven to be the most easy for the end user.

bedroesb: Out of experience people do not like to fill in the yaml ....

bedroesb: @jburel lets take this discussion to #597

martin-nc:
I see we haven't got a convention for times in the [style guide](https://rdmkit.elixir-europe.org/style_guide.html). Perhaps 09:00 is neater than 9h00? That's just personal preference. Also, do we need to repeat 'CET'? Can it just be on the last one when there are two times?

jburel:
cf. discussion above about schema. Event if schema.org is not in used for now, it is preferable to use 09:00 UTC+2 I guess in that case.
Adding an hyperlink to the time is also useful to point to https://www.timeanddate.com/worldclock/fixedtime.html?p1=3903&iso=2021-06-08T11:15:00 so people can find out the time for the timezone they are in.

martin-nc:
Actually, I was wrong about not having a convention. As the style guide says at the top, we follow the [European Commission style guide](https://wikis.ec.europa.eu/display/WEBGUIDE/07.+Dates%2C+numbers+and+symbols) (this project is funded by the EC). They have 09.00 (CET) as their convention. 

So I think we either follow that or, since our primary audience is people in Europe who work in bioinformatics and data management, ask the question,"What will this audience find most intuitive?" In this case we could follow the conventions already used in that world e.g.
SIB: https://www.sib.swiss/training/course/2021-06-genex-bgee-lec
EBI: https://www.ebi.ac.uk/training/events/guide-proteomics-data-analysis-using-uniprot-and-interpro/
where both use the format 09:00 - 13:00 CET/CEST/BST etc. 

Of course if we use if we use schema.org markup then that date format could be more machine-readable because the user won't necessarily see it. 

Personally I don't think it's necessary to provide a link to help convert time zones, but maybe that's just me. Again, this is a project funded by the European Commission so the target audience is predominantly European, and we are used to dealing with the time zones here.

bedroesb:
@martin-nc I updated the time format and added documentation in the contributors guide:

Screenshot from 2021-06-17 00-04-18

martin-nc (approved):
@bedroesb Looks okay to me now. Sorry for being slow - I'd forgotten the contentathon is next week!

@jburel Thanks for the input! Let's deal with the schema.org suggestion separately (I think @bedroesb has opened an issue). It has also been suggested we adopt [Bioschemas](https://bioschemas.org/) markup, but I'm less clear how we might do that. The types are pretty specialist.

jburel:
@bedroesb @martin-nc I registered for the event over a week ago but I have not received any information or link unless I missed something

jburel:
sorted

PR #612, opened 2021-06-22
optional_template_infrastructure_tool_assembly.svg

korbinib:
Although we have not really agreed on a template, I have been asked multiple times about this figure in a generic form. Would include here as a facultative add-on (not as a recommendation).


martin-nc:
@korbinib Looks like a good start. Did you use layers when creating it though? In Illustrator and an [online vector editor](https://drawsvg.org/) all the elements were in one layer. I find it a little easier to edit if there are separate, named layers, like "arrows", "circles", "text", "icons", at least at the start. Then you might find there is a better way to group things. It means you can lock some elements more easily whilst playing around with others. It also makes things easier to find if you want to do something to the layer an object is in. Maybe you did save in layers and that just hasn't come through?

I can add the layers if you haven't already, but that can be after this PR, as a second iteration with the icons.

korbinib:
@bedroesb @martin-nc  Updated with font awesome icons (only CC-BY solid at the moment) please have a look and let me know how you think this works for you.

bedroesb (approved):

PR #624, opened 2021-06-24
fontawesome

korbinib:
change if we license fontawesome pro

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -2,11 +2,15 @@
 
 The RDMkit makes all of their materials publicly available under [open source][osi] licenses.
 
-The process documents and data are made available under a CC-BY license.
+The process documents, icons and data are made available under a CC-BY license.
 Software are made available under an MIT license.
 
 The license text listed below (describing both CC-BY and MIT licenses as well as their usage in the RDMkit) is reused under a CC-BY license from The Turing Way community materials.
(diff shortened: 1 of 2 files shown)


LICENSE
```diff
@@ -2,12 +2,16 @@
...
-The process documents and data are made available under a CC-BY license.
+The process documents, icons and data are made available under a CC-BY license.
 Software are made available under an MIT license.
 
 The license text listed below (describing both CC-BY and MIT licenses as well as their usage in the RDMkit) is reused under a CC-BY license from The Turing Way community materials.
 (Specifically from the [The Turing Way](https://github.com/alan-turing-institute/the-turing-way/blob/master/LICENSE.md)).
 
+## Icons
+
+Icosn used in this repository are licensed under [Creative Commons Attribution 4.0 International License][cc-by-human] from [fontawesome](https://fontawesome.com/license/free)
```
martin-nc, line 13: A little typo: "Icosn" instead of "Icons". I will ask for a Pro license today (I need to request the purchase), but I think we need to make sure the icons suit the life cycle diagram as well before we fully adopt FontAwesome.

bedroesb:
@korbinib Do we have to change here something if we stick with the noun project?

korbinib:
decided for noun icons instead

PR #629, opened 2021-06-25
Top nav improvements + other small changes

bedroesb:
This PR will bring
- [x] a home button to the top navigation
- [x] take away the feedback button on the top of the website
- [x] search icon next to the search bar so people can spot it quicker
- [x] better link towards the email address in the footer
- [x] info about the country column that is added to the tools table
- [x] moving the social icons in the contributors panels to a footer section
- [x] making the search icon part of the search bar

You can find here a [preview](https://bedroesb.github.io/rdmkit/)
Screenshot from 2021-06-25 22-34-17

<hr>

Screenshot from 2021-06-25 22-37-28


martin-nc:
Would the search icon stand out more against the white of the search box? It's a little lost against the black, I find. If it's a button you can always style it so it joins the search box (give it a white background, and square rather than rounded edges on the left). Either that or try something similar to what I did for the ELIXIR search, but this time with a maroon rather than orange background.
Screenshot 2021-06-25 at 22 06 49

bedroesb:
@martin-nc It would be nice if it is within the whitespace just as an icon, but I did not immediately manage Something that looks like a button would not be a good idea, since it is not a button ... :)

martin-nc:
@bedroesb Ah, I didn't look to see if it was a button or not. Maybe adding a button would make the search stand out more? As for putting the search icon in the search box, I'd put it in the left to avoid giving the impression that you can click on it. Could you use the ::before pseudo-element, like in [this article](https://www.denisbouquet.com/how-to-use-font-awesome-icons-in-css/)? Only, of course, you wouldn't be using FontAwesome. You'd be using an SVG from the Noun Project ;o). 

A [more fashionable approach](https://css-tricks.com/icon-fonts-vs-svg/) is to use an inline SVG (e.g. Google, Stack Overflow). You could wrap the search input box in a div that is relatively positioned. Then put the SVG code for the search icon in the div and absolutely position it in the search box. Something like:
```
<div class="position-relative">
  <input type="text" placeholder="Search..." aria-label="Search" />
  <svg class="position-absolute" aria-hidden="true">
   ...
  </svg>
</div>
```
Stack Overflow does this. Or perhaps you've tried that.

bedroesb:
@martin-nc I added the search icon with font awesome for now ... I see in your linked website that svg has some serious benefits, but the fact that I do not have to download an svg and link towards it makes it a lot quicker to implement it this way:

Screenshot from 2021-06-30 18-39-29

I alsoa dded some extra margin in between the navigation titles in mobile mode

martin-nc (approved):
@bedroesb Looks good to me!

PR #631, opened 2021-06-29
Replacing tags with page titles

bedroesb:
- [x] adding mechanics to read the tag from page metadata -> tags.yaml is not needed anymore, one file less to keep up to date
- [x] adding tags to all the pages (page_tag attribute)
- [x] replacing the word tags with 'related pages'
- [x] page-metadata include at the end of the page
- [x] updating the documentation
- [x] adapting the tools table conversion script so it can check for tags in the page metadata??**
- [x] better styling for the table needed?  

[Preview](https://bedroesb.github.io/rdmkit/index.html) 

** QUESTION: do we need a controlled tags vocabulary or do we just not show them if they do not have a linked page. Advantage is that people can start adding tools with tags that do not yet have a page on RDMkit. The disadvantage is that we will not spot if a tool has for example the tag 'humandata' without space, and it does not show up there ...

Screenshot from 2021-06-29 18-19-30
Screenshot from 2021-06-29 18-19-23


floradanna:
1 page = 1 tag 

By looking at the image above, I though for a second that "Related pages" and "Contributors" were related/linked. Maybe "Related pages" and "Contributors" should not be shown together with the same layout. They could have different places in the page.
"Related pages" brings the reader out of the current page.
"Contributors" refers to authors of the current page.

smza:
> 1 page = 1 tag
> 
> By looking at the image above, I though for a second that "Related pages" and "Contributors" were related/linked. Maybe "Related pages" and "Contributors" should not be shown together with the same layout. They could have different places in the page.
> "Related pages" brings the reader out of the current page.
> "Contributors" refers to authors of the current page.

Flora, I think the only thing different from what we have currently is that Bert had replaced the "Tags" by "Related pages". Otherwise this is how the information is displayed at the bottom of the pages currently. Maybe we could have separate panels Or we could move contributors to the top of the page.

bedroesb:
I wonder if we need the current page title within the tags:

Screenshot from 2021-07-01 10-16-12

In this case I could leave out `researcher`

bedroesb:
Screenshot from 2021-07-01 12-01-45

@floradanna this si when they are seperate

bedroesb:
We can later brainstorm about the order we want to give things on the page: Related pages, Tools and resources, training, fairsharing, DSW, ...

PR #634, opened 2021-06-30
adding more country tags to DSW

korbinib:
@bedroesb This might be a potential manifestation of earlier discussed problem on multiple instances of one tool with different accessibility.


bedroesb (approved):

MarekSuchanek:
Cannot this cause a confusion – that DSW is used *only* in CZ, NL, FI, NO, SE?

korbinib:
@MarekSuchanek Yes. I think we need a system to track tools vs their (local) instances. (and a good visual representation) This holds true for other tools as well e.g. galaxy, seek, etc. I would see this as a temporary state, hope we can fix this in general later this year.

bedroesb:
This topic is toughly discussed in the Biohackathon, and we decided to not list country specific tools (including DSW) in the main toolstable. Countries will have to list their DSW instance on there national resource page after #749 is merged.

PR #635, opened 2021-06-30
Minor fixes for the style guide

martin-nc:
The section on references had a stray angle bracket and an unintended bullet point. I've also added the fact that the plural of "tool assembly" is "tool assemblies", in case someone was tempted to write "tools assemblies".


bedroesb:
@martin-nc do we also mention something about the noun project?

martin-nc:
@bedroesb Oh yes - I've added that now.

bedroesb (approved):

PR #639, opened 2021-07-02
Registry badges update

bedroesb:
This part is a start to make the linking of tools to other registries more clear. The current look in the screenshot is not final and I need help in deciding what will be the text in the badge, and what will be the text when you hover over it. 

[Preview website](https://bedroesb.github.io/rdmkit/all_tools_and_resources.html)

Screenshot from 2021-07-02 15-56-32

@lauportell @korbinib @martin-nc 

TO DO:
- [x] 1. bio.tools has a lowercase 'b' at the beginning. Of course that's not usual in a proper noun, but there we are.
- [x] 2. I would make all the labels consistent, so they just name the resource: 'FAIRsharing', 'bio.tools' and 'TeSS'. Then when you hover over them you'd get: 'Find related policies and standards' (FAIRsharing), ' Find a scientific and technical description of the resource' (bio.tools) and 'Find training for this resource' (TeSS).
- [x] 3. I'm not sure about the label colours. I wouldn't know which orange applied to which resource, and orange is a warm colour so it jumps out. I'd be worried it jumped out more than the actual resource/tool name! Maybe they should all be the same colour, but a cooler one. Would the link blue work as a background? Or the dark grey or dark blue we have in the style guide (with contrast reversed)?
- [ ] 4. I'd make the [table heading sticky](https://css-tricks.com/position-sticky-and-table-headers/).
- [x] 5. I find the hover state on the rows distracting. I think the table rows are big enough for you to understand which one you are on. Maybe that's just my taste, though.

### DIFF
--- a/_includes/page-metadata.html
+++ b/_includes/page-metadata.html
@@ -1,6 +1,6 @@
 {%- if page.tags != null %}
 <div class="meta-line p-3 mt-3">
-    <b>Pages that link to this page: </b>
+    <b>Related pages: </b>
     {%- for tag in page.tags | sort %}
     {%- assign related_page = site.pages | where:"page_tag",tag | first %}
     <a href="{{ related_page.url }}" class="meta-obj"
--- a/_includes/toollist.html
+++ b/_includes/toollist.html
@@ -10,7 +10,7 @@
 The tag "{{ include.tag }}" has not yet related tools or resources.
 {%- else%}
 <a class="visually-hidden-focusable" href='#skip-tool-table'>Skip tool table</a>
-<table id="tooltable" class="table table-hover display my-5">
+<table id="tooltable" class="table display my-5">
     <thead>
         <tr>
             <th>Tool or resource {%- if include.tag or include.tag2 -%}
@@ -20,7 +20,11 @@ The tag "{{ include.tag }}" has not yet related tools or resources.
             </th>
             <th>Description</th>
             <th>Related pages</th>
-            <th>Registry</th>
+            <th>Registry {%- if include.tag or include.tag2 -%}
+                <a data-bs-toggle="tooltip" data-bs-original-title="Links to related information in ELIXIR registries: related policies and standards in FAIRsharing, scientific and technical descriptions of the resource in bio.tools, and related training in TeSS.">
+                    <i class="fas fa-info-circle" style="color:var(--main-text-color);"></i>
+                </a>{%- endif %}
+            </th>
         </tr>
     </thead>
     <tbody>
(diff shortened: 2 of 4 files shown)


martin-nc:
Here are a few thoughts:
1. bio.tools has a lowercase 'b' at the beginning. Of course that's not usual in a proper noun, but there we are.
2. I would make all the labels consistent, so they just name the resource: 'FAIRsharing', 'bio.tools' and 'TeSS'. Then when you hover over them you'd get: 'Find related policies and standards' (FAIRsharing), ' Find a scientific and technical description of the resource' (bio.tools) and 'Find training for this resource' (TeSS).
3. I'm not sure about the label colours. I wouldn't know which orange applied to which resource, and orange is a warm colour so it jumps out. I'd be worried it jumped out more than the actual resource/tool name! Maybe they should all be the same colour, but a cooler one. Would the link blue work as a background? Or the dark grey or dark blue we have in the style guide (with contrast reversed)?
4. I'd make the [table heading sticky](https://css-tricks.com/position-sticky-and-table-headers/).
5. I find the hover state on the rows distracting. I think the table rows are big enough for you to understand which one you are on. Maybe that's just my taste, though.

bedroesb:
@martin-nc Thanks martin for the feedback, one thing I was wondering, we now have the info icon next the the Tool or resource column, can't we use the same to explain the registry column?

martin-nc:
@bedroesb Yes, that's an idea. Do you mean instead of hovering over the registry name, or instead of a footnote, or instead of both?

The Tool or resource icon isn't in your screenshot, though, for some reason.

bedroesb:
@martin-nc true because the table gets explained in the text above the all tools and resources page. The info icon only gets used in the tool tables on the other pages.

bedroesb:
@martin-nc the dticky header doesn't seem to work in combination with `overflow-x: auto;` and the solutions are not [straight forward ](https://uxdesign.cc/position-stuck-96c9f55d9526)

lauportell:
Hi Bert, sorry I didn't say anything about this before. I agree with all Martin's suggetions and I think it looks very good right now. I don't think they need different colours and it is perfect that when you hover over you see the explaination of what you will find when you click. Thank you!!

martin-nc:
@bedroesb Ah, okay - I don't think it's worth going to that length. Maybe leave them non-sticky for now.

martin-nc:
Actually, I've just had another thought (sorry!). Would it look better to make the first column bold and decrease the font size of the description to .9em or so, just to decrease the length of the table and make the tool name stand out more? I guess something like (untested!):
```
td:nth-child(1) a {font-weight: bold;}
td:nth-child(2) {font-size: .9em;}
```

bedroesb:
I think it is an improvement, you can see the update in the preview. Thanks @martin-nc !

martin-nc (approved):

Issue #594, opened 2021-06-10
Icons toolbox


korbinib:
Outgoing from #591 is there a need for an icons toolbox to make the illustrations more consistent?

Would like to suggest to use https://github.com/ebiwd/EBI-Icon-fonts and the inbuild https://fontawesome.com/ as a starting point, but we might have to evaluate licensing issues.

floradanna:
How about using the icons we have in the RDMkit data life cycle image? Probably we will need few additional ones as well.

martin-nc:
Unfortunately there isn't a standard icon set in ELIXIR at the moment. This is going to be discussed later on in the summer by at least Xenia, Erin and me. Anyone else is free to chip in with ideas. 

I use Font Awesome on the ELIXIR site sometimes. For now, I agree with @floradanna that we should use the icons in the RDM cycle where relevant. The EBI set is also a possibility, but we use some icons as symbols of the Platforms, so we have to be careful not to use those! We use the 'toolkit' icon to represent the Tools Platform, for example, so clashes like this might be a problem.

I do think we need to tell contributors that illustrations may need to be edited so that they conform to the brand colours and icon styles. It's best to start this process early, so we don't have to redraw images already published on the site!

korbinib:
> The EBI set is also a possibility, but we use some icons as symbols of the Platforms, so we have to be careful not to use those! 

Or the other way round it could be also a way to represent visual linkage between the concepts/requirments in the RDMkit and the platforms which severe the demands.

> I do think we need to tell contributors that illustrations may need to be edited so that they conform to the brand colours and icon styles. It's best to start this process early, so we don't have to redraw images already published on the site!

+1 agreed any chance we have a guide for this in place within the next 2 weeks?

bedroesb:
We should try to have this in place as fast as possible indeed, I will lookup which icon set I used for the data life cycle wheel

bedroesb:
Since it is a long time a go I made the current data lifecycle wheel, I had to dig, and now I remember that I used multiple sets in the https://thenounproject.com/ . I did not find one icon sets that contained all the icons we needed... This is for example the [reuse icon](https://thenounproject.com/deemakdaksina/collection/nature-ecology-solid/?i=1754090)  and this is the [flask icon](https://thenounproject.com/search/?q=flask&i=1826710)

martin-nc:
[...]

PR #649, opened 2021-07-08
noun project icons for tool assemblies

korbinib:
We have started to use a common icon set from the [noun project ](https://thenounproject.com/ELIXIRCommunications/kit/rdmkit/). (See #635 #594 )

In connection to this I have modified all tool assembly illustrations using the previous EBI icon font accordingly.

@minnami @siirif - Could you please check if this looks OK for you?


bedroesb:
@korbinib looks good. I assume the plan icon is centered in the circle, yet it looks not centered probably because of the page being the main part of the icon. But this is only a small detail :)

korbinib:
Yes, it's indeed centred . If we want it optically centred we would have to decrease the size of the icon to fit it in the circle. Can do that if you want.

bedroesb:
@martin-nc do you think it is needed to optically center ? :)

martin-nc:
FWIW I'd optically centre them. The 'share' icon looks like it needs nudging left a bit as well, at least to my eyes.

korbinib:
@martin-nc @bedroesb hope this looks OK now?

bedroesb (approved):

PR #654, opened 2021-07-13
Update fr_ifb_assembly.md

OlivierCollin:
First version of IFB tool assembly. I also have a svg file but I can not push it in the image directory.


korbinib:
Thank you @OlivierCollin I have added some minor link and style changes. If you could send me your SVG, I can add it to the branch and to this pull request.

korbinib (approved):

bedroesb:
@OlivierCollin do you think the page is ready to be linked in the sidebar? I can add it. The only thing that is missing is the image.

OlivierCollin:
@bedroesb Hi Bert, the initial svg file has been sent to Korbinian for review and editing.

bedroesb:
@OlivierCollin that is great!

korbinib:
Since this is distributed across different repos and branches now. I have just committed the SVG  [e1b5661](https://github.com/elixir-europe/rdmkit/commit/e1b56617e420ccea27569caa8d3397a46d6e290d).  I see the IFB tools still have to be included in the tools table. Will edit #645

bedroesb:
@korbinib I rather have that you update the spreadsheet, that makes things less confusing :)

OlivierCollin:
I made some IFB tag addition to the list.

OlivierCollin:
I don't know if we should only add node tags (e.g. IFB) to all tools or focus only on specific tools but in this case it means we are using some country tag. I am afraid that there will be a lot of tags for some tools. For example, there is Galaxy for almost every node, or in the case of Conda/Bioconda it will be almost everywhere.

floradanna:
@korbinib What's the policy for instances of a software (Galaxy, DSW etc..) ? For @OlivierCollin

OlivierCollin:
This will be the same for other IT technologies : Conda, Openstack, iRODS, etc. How far should we go ?

OlivierCollin:
My concern is that the spreadsheet might become unmanageable as such and that errors might happen during the edition process with too many tags.  However this could be quite interesting to see the extend and adoption of a technology across Elixir nodes.

korbinib:
@OlivierCollin  My view on this as long as it helps someone to use it or to build something with this as an inspiration I would list the tool. As a hard criterion, I feel the tool should have context given in the text. Happy to hear other opinions. - @floradanna ?

OlivierCollin:
@korbinib @floradanna I completely agree with you, giving an overview of the Elixir technological portfolio is quite interesting for many. I am not sure to get the idea of the hard criterion. It will be the fact that, in a tool assembly, the tool is associated to some description giving a hint of its usage and interest ?

korbinib:
@OlivierCollin Sorry about the ambiguity.  I meant: There is maybe not so much use in tagging OpenStack for the tools table of a tool assembly if it is not mentioned in the text that the infrastructure X is based on it.

OlivierCollin:
@korbinib Yes, this was exactly the problem I encountered. I tagged Openstack and started to wonder if it was a good move, since, having tagged Openstack, I was asking myself about all other technologies like conda, docker, singularity. These technologies are quite ubiquitous on many bioinformatics infrastructures and there is not a lot added value to mention it.

korbinib:
@OlivierCollin I think this is also a question of scoping. We are (mainly) targeting researchers and data stewards - they might be more interested in the capabilities and will pitch this to someone who could deploy it then for them.

PR #664, opened 2021-07-20
First DS Wizard deep link proposal

rwwh:


DanFaria (changes requested):
@rwwh maybe it should be stated that the DSW link requires a DSW account to access it.

rwwh:
The deep link should *not* require a DSW account, @DanFaria. If it does now, that is an error somewhere. Did you test it?

DanFaria:
I did test it

image

rwwh:
Thanks, and sorry. This will need a fix on the DSW side (probably a configuration flag in the "researchers" instance). The patch should be unaffected.

DanFaria:
So, we put this patch on hold until you sort things out on the DSW side?

rwwh:
I'm working on it. 

I think it would be good to discuss any other aspects of the patch already, like the general structure of the referral, and how this fits into the future view on page templates. That was the primary reason for me to submit this PR.

DanFaria:
Yeah, sorry for leaving that part out. I do agree with your proposed structure---I was about to accept the pull request when I thought to check the link first :-)
I would like to bring this up in the next editorial board meeting, since this will require updating page templates, but I don't anticipate any issues.

MarekSuchanek:
There was a misconfiguration for public knowledge model previews... it works now.

DanFaria:
I confirm it is working!

PR #665, opened 2021-07-22
Improve DSW descriptions

MarekSuchanek:
The descriptions of DSW were rather brief (e.g. title "DS-Wizard", description "Data Stewardship Wizard"). It was also not clear in what order are the tools listed - any other ordering is OK as well, just state what is the order. Finally, links for some tools headed to a landing page but for other to a login page or directly to a specific instance.


bedroesb (approved):

bedroesb:
Thank you!

PR #690, opened 2021-09-03
fix registry zenodo in training

floradanna:
fix registry zenodo in training


floradanna:
There is a mistake in all training resources table. The following registries should be Zenodo, not TeSS.
@bedroesb how can it be fixed?

Screenshot 2021-09-03 at 21 47 36

bedroesb (approved):

bedroesb:
It was a little bug in the script :)

PR #699, opened 2021-09-14
Contributor cards on page level + better buttons

bedroesb:
Instead of being redirected to the github page of a person by clicking on its name at the bottom of a page, a contributor card is opened with picture and socials.

Screenshot from 2021-09-14 18-26-18

I also use this PR to replace all custom styling concerning buttons with bootstrap classes.


bedroesb:
@martin-nc I am aware that the links in these cards can not be accessed through tab (for accessibility), but this is a limitation of [popovers](https://getbootstrap.com/docs/5.1/components/popovers/#usage). I hope this is not a huge problem?

martin-nc:
@bedroesb Would a dropdown/dropup be a more accessible option?

bedroesb:
I don't think this will behave nicely when there are multiple rows. I also don't think we hide crucial information away from visibility impared people.

Screenshot from 2021-09-14 19-47-18

A live preview can be found [here](https://bedroesb.github.io/rdmkit/data_steward_policy.html)

martin-nc:
I hadn't thought about multiple rows. The only thing now is that it doesn't work in Safari. You click the button and nothing happens. Error in console: "TypeError: undefined is not an object (evaluating 'this._element.closest')." Perhaps it would help to load JQuery at the bottom of the page rather than in the `<head>`?

Screenshot 2021-09-15 at 10 12 26

bedroesb:
@martin-nc I did some fixes, can you check if it now throws errors/works?

martin-nc (approved):
Works fine now!

PR #702, opened 2021-09-20
Support for news + events page

bedroesb:
This PR will add the two latest news items to the frontpage + adds a dedicated events page with all past en upcoming events

Screenshot from 2021-09-20 13-45-26

### DIFF
--- a/_data/news.yml
+++ b/_data/news.yml
@@ -1,0 +1,4 @@
+- name: Open Science FAIR
+  date: 2021-09-20
+  description: Test
+
--- a/_data/sidebars/about.yml
+++ b/_data/sidebars/about.yml
@@ -9,6 +9,10 @@ subitems:
         url: /editorial_board.html
       - title: Support
         url: /support.html
+  - title: News
+    url: /news.html
+  - title: Events
+    url: /events.html
   - title: Media kit
     url: /media_kit.html
   - title: Outreach
--- a/_includes/events.html
+++ b/_includes/events.html
@@ -1,6 +1,6 @@
 
-<div class="events">
-    <h3 class="mt-5">Upcoming events</h3>
+<div class="events upcoming-events">
+  <h2 class="mt-5 text-dark">Upcoming events</h2>
     {%- assign events = site.data.events %}
     <ul class="list-unstyled mt-3">
         {%- for event in events %}
@@ -12,11 +12,12 @@
           <p class="text-muted mb-0"><i class="fas fa-map-marker-alt me-2"></i>{{ event.location }}</p> 
           {%- endif %}
           {%- if event.description %}
-          <p class="text-muted">{{ event.description }}</p>
+          <p>{{ event.description }}</p>
           {%- endif %}
         </li>
         {%- endfor %}
     </ul>
+    <small>An overview of all our events can be fount on the <a href="/events.html">events page</a>.</small>
 </div>
 <script>
   $(document).ready(function() {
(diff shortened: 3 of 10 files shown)


_data/news.yml
```diff
@@ -0,0 +1,4 @@
+- name: Open Science FAIR
+  date: 2021-09-20
+  description: Test
```
bedroesb, line 3: @korbinib @lauportell @floradanna @smza @wittigue 

Could someone help me out with Adding some news items by commenting or changing this yaml file? Thanks!

korbinib: multi day events don't have an end-date, do they?

bedroesb: Well I wanted to not let it look like an event ... Because it is a news item :)

bedroesb: The date should actually be the date the news item is added

bedroesb: So all the info you want to include has to be done through the description attribute.

lauportell: yes, will add the ones I know

martin-nc:
Here are a few comments from me:

1. On the design of the home page boxes: I wonder if the red and black for each event and news item are a bit distracting. They're pretty strong colours. Maybe the page sections would be clearer if the whole news and event boxes had the line by them, and only the heading was red.
Screenshot 2021-09-20 at 15 58 05

2. Do we need a contact email for each event, or at least a clear indication who can join and how? The Machine- Readable Metadata session has the Zoom link (is that safe?) and says 'Get involved!' but is that for people working within the project, or anyone?
3. There's a typo in the sentence under the events block ('fount' rather than 'found'). Maybe just put 'See the events page for past events'. Presumably all upcoming events will be on the home page, so there is no point going to the events page to see more of them.

bedroesb:
@martin-nc I agree with all of them :)

KarelBerka:
ad 2.  it is open to anyone and hence Zoom link is prepared only for that occasion.

korbinib:
> ad 2. it is open to anyone and hence Zoom link is prepared only for that occasion.

We could use a Zoom registration forms in future to reduce the likelihood of zoombombing.

floradanna:
My personal opinion would be to remove the zoom link from the event in the front page. I am not an expert in security, but currently anyone can click the zoom link and start the meeting right away. I am not sure this is acceptable.

We can still the announcement of the events as something that DM experts will do for RDMkit and maybe put the data-management-coordinators-owner@elixir-europe.org email as contact, in case someone is interested to join.

floradanna:
For next time we should definitively think of something for security, if we still want the zoom link there. But right now, I strongly recommend to remove the open zoom link

martin-nc:
I've asked David at the Hub. He has done most with Zoom and online meeting software. Like Korbinian, he recommends the Zoom registration form, because you can set up reminder emails and it has a link to add the event to your calendar.

Nazeeefa:
Hi! This is very nice, thanks.
Did I understand correctly, there will be two pages: one for events and one for news?
Also, by "adds a dedicated events page with all past en upcoming events", do you mean there is a plan to create a new page (and place it in a top menu) that can also be accessed via 'Upcoming events' title (serving as a button) mentioned on home page?

bedroesb:
@Nazeeefa Only upcoming events are shown at the frontpage at the moment. With this pullrequest I will indeed add a dedicated events page where I will point at with a little sentence + link saying that if you want to also see paste events you can click here. These events and newspage will fine their place in the about section of the website.

bedroesb:
@martin-nc I tried your suggestion but it did not fit super well with the other subtitles below the news section and it felt a bit out of place. What do you think of this:


Screenshot from 2021-10-06 14-29-20

Or this one:

Screenshot from 2021-10-06 14-35-21

Issue #641, opened 2021-07-05
New page: proteomics domain page


miantur:
Dear RDMkit team,

I would like to suggest a new page for the proteomics domain. This domain-specific page is still missing. So, this could close this gap. I have already prepared a first draft for the content. Please feel free to comment it. I hope that this fits into the "your domain" section of the RDMkit. I am looking forward to read your comments/suggestions.

Thanks & best regards,
Michael  

**What topic do you wish to add?**
I would like to suggest a new page for the proteomics domain.

**Are there existing pages in the RDMkit website related to the requested page?**
In the section "your domain" other domains are already described. Proteomics is still missing.

**Resources**
A first draft for the suggested page can be found and commented here:
https://docs.google.com/document/d/1P9t-mXSwjP65X5k9Jgqeu0JPURBk_jcdgoTM0DSbvlk/edit#heading=h.lqxgf6rdmm1z

**Context**
This request addresses the description of domain-specific RDM aspects within the proteomics domain.

PR #707, opened 2021-10-01
new domain page "Proteomics"

wittigue:
#641


bedroesb (approved):

korbinib (approved):
I feel this covers the central parts.

Would be interesting to hear if there is best practises to complement MIAPE for documenting experimental steps before proteomics sample preparation.

PR #709, opened 2021-10-04
Extract DSW deep link prefix to var

MarekSuchanek:
With this, we can easily change the links in the future (e.g. when we have the intermediary "pick-your-DSW-instance/KM" page).

Jekyll/Liquid filter would be nicer and allow parameters but I think that is [not possible](https://docs.github.com/en/pages/setting-up-a-github-pages-site-with-jekyll/about-github-pages-and-jekyll#plugins) to get working with GitHub Pages (it would need to be a Ruby plugin). If it is possible, I can create the plugin...

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -25,6 +25,9 @@ port: 4000
 gtag: G-RXQ55EFTTH
 # Google analytics tag
 
+dsw_deep_link_prefix: https://researchers.ds-wizard.org/knowledge-models/dsw:root:latest/preview?questionUuid=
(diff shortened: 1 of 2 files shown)


bedroesb:
True ... one of those limitations of Github pages, I think the solution you offer is not perfect, but the best one for now.

floradanna (approved):

bedroesb (approved):

bedroesb:
Should we not add something in our editorial guide about DSW @floradanna ?

PR #749, opened 2021-11-11
Start for national resource pages

bedroesb:


bedroesb:
@korbinib be careful with the indentations of the resources, most of them had some mistakes (causing the error). I would also not repeat the country name in every heading. You are on the Norway page so I think repeating this is not needed. I don't know if we already discussed which headings we would take, since I see you changed the ones we've put in the template:

## Introduction 


## Funders

## Regulations

## Domain-specific infrastructures/resources

korbinib:
How fixed are the headings, I think we also had `Institutions` at some point? This is a quick draft with work still in progress.

floradanna:
I think the headings are not fixed yet. I do agree that repeating the name of the country in every heading is not very useful, though. We could propose and discuss changes in the template in an issue or something.

I would be in favour of keeping the national pages template a bit flexible, since every country might have different needs. 

However, I remember that we discussed about having institution-level information during BH. This option seemed not really feasible for countries with numerous universities and institutions. But we can review this topic again, if needed.

bedroesb:
all things of the RDMkit meeting are implemented: 

- We use the general tools and resources list as a CV for the local flavors/instances of a tool
- This is to be used in a new field in the resources listed on a country page
- The all tools and resources page will list all tools including the national ones, and it is OK they appear "double"
- Those local instances/flavors/countryfocused tools will be listed on the country page
- We keep the national resources at the bottom of the table on the respective page and reveal them with a button
- The general tool in the tools table will have a visible "thing" that tells them there are national flavors at the bottom of the page for this tool

PR #760, opened 2021-11-19
Upgrades to the tool assembly index page

### LINKED ISSUE #742: Implement a filtering system for tool assembly (by country, by domain) or a search function to search on description.

### LINKED ISSUE #743: Implement API to fetch the country name corresponding to the ISO country code.

bedroesb:
- [x] Filtering based on affiliation (closes #742 
- [x] Adding search bar to search on title/description
- [x] Equal size of affiliations/audiences buttons 
- [x] Country names instead of country codes are displayed
- [x] Update country list using API (closes #743)

image

Live preview:

https://bedroesb.github.io/rdmkit/tool_assembly.html

### DIFF
--- a/.github/workflows/list_conversion.yml
+++ b/.github/workflows/list_conversion.yml
@@ -23,6 +23,9 @@ jobs:
       - name: Run tool table 2 yaml
         run: |
           python var/conversions.py --reg --username ${{ secrets.FAIRSHARING_USERNAME }} --password "${{ secrets.FAIRSHARING_PASSWORD }}"
+      - name: Fetch country names with corresponding iso
+        run: |
+          python var/country_list.py "
       - name: Show differences
         run: 'git diff --stat'
       - name: Create Pull Request
--- a/.github/workflows/list_conversion.yml
+++ b/.github/workflows/list_conversion.yml
@@ -25,7 +25,7 @@ jobs:
           python var/conversions.py --reg --username ${{ secrets.FAIRSHARING_USERNAME }} --password "${{ secrets.FAIRSHARING_PASSWORD }}"
       - name: Fetch country names with corresponding iso
         run: |
-          python var/country_list.py "
+          python var/country_list.py
       - name: Show differences
         run: 'git diff --stat'
       - name: Create Pull Request
--- a/_data/countries.yml
+++ b/_data/countries.yml
@@ -1,0 +1,250 @@
+AD: Andorra
+AE: United Arab Emirates
+AF: Afghanistan
+AG: Antigua and Barbuda
+AI: Anguilla
+AL: Albania
+AM: Armenia
+AO: Angola
+AQ: Antarctica
+AR: Argentina
+AS: American Samoa
+AT: Austria
+AU: Australia
+AW: Aruba
+AX: "\xC5land Islands"
+AZ: Azerbaijan
+BA: Bosnia and Herzegovina
+BB: Barbados
+BD: Bangladesh
+BE: Belgium
+BF: Burkina Faso
+BG: Bulgaria
+BH: Bahrain
+BI: Burundi
+BJ: Benin
+BL: "Saint Barth\xE9lemy"
+BM: Bermuda
+BN: Brunei Darussalam
+BO: Bolivia (Plurinational State of)
+BQ: Bonaire, Sint Eustatius and Saba
+BR: Brazil
+BS: Bahamas
+BT: Bhutan
+BV: Bouvet Island
+BW: Botswana
+BY: Belarus
+BZ: Belize
+CA: Canada
+CC: Cocos (Keeling) Islands
+CD: Congo (Democratic Republic of the)
+CF: Central African Republic
+CG: Congo
+CH: Switzerland
+CI: Ivory Coast
+CK: Cook Islands
+CL: Chile
+CM: Cameroon
+CN: China
+CO: Colombia
+CR: Costa Rica
+CU: Cuba
+CV: Cabo Verde
+CW: "Cura\xE7ao"
+CX: Christmas Island
+CY: Cyprus
+CZ: Czech Republic
+DE: Germany
+DJ: Djibouti
+DK: Denmark
+DM: Dominica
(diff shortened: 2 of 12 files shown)


martin-nc:
I think it's much clearer than before! Here are some thoughts (as ever, feel free to ignore!):
1. Is there meant to be a red line next to the opening paragraph? Perhaps that's not needed and the text can be a normal paragraph?

2. Can I suggest an edit of the opening paragraph? I found the second sentence hard to understand because of the two "that"s.

> A Tool Assembly is a combination of tools that can help with data management tasks. The tools are assembled by user communities, but the assemblies can often be used by people outside those communities. The assemblies are aimed at users in a specific location and/or for users within a specific domain.


3. (Minor point) I wasn't sure what "affiliation" meant at first in the filter.  I don't think of software having affiliations, but maybe that's just me. Would "origin" or "source" be any better? I wasn't sure what "Affiliations: ALL" meant, too. Does it mean that the assembly is not associated with any country or project? 

4. Does "Audience" mean only the people from that country can use the assembly, or can people outside use it as well? The NeLS entry makes this clear. It says it is aimed at "researchers in Norway and their collaborators". The IFB entry, though,  says the audience is France but the blurb says the assembly supports "researchers in Life Sciences". Perhaps the blurb for each assembly should make it clear who the audience is, or perhaps the audience should be a text field so that it can explain the nuances better?

5. Will the flags eventually link to the country pages? I'm asking because there's a little inconsistency in the UI. I can click on the ELIXIR and CSC buttons, but I can't click on the flag or 'ALL' buttons, even though their design suggests I can.

6. When I filter for Euro BioImaging I get "Plant Genomics", but there's no Euro BioImaging tag on that. There is on the other two results. Has the filter picked Plant Genomics up because it has no affiliation?

7. I find it a little distracting that the same fields in the boxes appear at different heights within each box. So if you're on a desktop and you scan horizontally across two boxes, the 'Your tasks' heading doesn't horizontally align across the boxes (just sometimes). This makes the pages more difficult to scan. Perhaps the card blurb can have a `min-height`, so that the "Your tasks" heading is more likely to appear in the same location? I appreciate it's difficult to get uniformity across all cards, but perhaps this will help.
tasks-aligned

korbinib:
This is really nice. Some additional thoughts from me on this:

1. Should the filter be for affiliations or better be for the audience. 
2. Is there any way we can make 'ALL' catch all (also maybe there is a better name, but I cannot think of any at the moment)
3. ELIXIR Europe might be misleading for the filter(?)
4. Aligning the 'Your tasks' might be nicer indeed optically

korbinib (approved):

bedroesb:
@martin-nc 

> Is there meant to be a red line next to the opening paragraph? Perhaps that's not needed and the text can be a normal paragraph?

Took it away
> Can I suggest an edit of the opening paragraph? I found the second sentence hard to understand because of the two "that"s.
Done

> (Minor point) I wasn't sure what "affiliation" meant at first in the filter. I don't think of software having affiliations, but maybe that's just me. Would "origin" or "source" be any better? I wasn't sure what "Affiliations: ALL" meant, too. Does it mean that the assembly is not associated with any country or project?

You are right, affiliation all doesn't make any sense. About the naming I guess we have to discuss it tomorrow in the editorial meeting

> Does "Audience" mean only the people from that country can use the assembly, or can people outside use it as well? The NeLS entry makes this clear. It says it is aimed at "researchers in Norway and their collaborators". The IFB entry, though, says the audience is France but the blurb says the assembly supports "researchers in Life Sciences". Perhaps the blurb for each assembly should make it clear who the audience is, or perhaps the audience should be a text field so that it can explain the nuances better?

To be discussed

> Will the flags eventually link to the country pages? I'm asking because there's a little inconsistency in the UI. I can click on the ELIXIR and CSC buttons, but I can't click on the flag or 'ALL' buttons, even though their design suggests I can.

The country flags will in the future link to the country pages on RDMkit, yes

> When I filter for Euro BioImaging I get "Plant Genomics", but there's no Euro BioImaging tag on that. There is on the other two results. Has the filter picked Plant Genomics up because it has no affiliation?

This was a bug and is fixed now

> I find it a little distracting that the same fields in the boxes appear at different heights within each box. So if you're on a desktop and you scan horizontally across two boxes, the 'Your tasks' heading doesn't horizontally align across the boxes (just sometimes). This makes the pages more difficult to scan. Perhaps the card blurb can have a min-height, so that the "Your tasks" heading is more likely to appear in the same location? I appreciate it's difficult to get uniformity across all cards, but perhaps this will help.

I don't always like min-heights because the height changes quite a lot at certain breakpoints. This would mean I always have to take the tallest height and this would not be space efficient in most wide breakpoint ...

bedroesb:
@korbinib 

> Should the filter be for affiliations or better be for the audience.

Is something I was wondering too, something to be discussed

> Is there any way we can make 'ALL' catch all (also maybe there is a better name, but I cannot think of any at the moment)

Since ALL as affiliation does not make any sense, I took it out of omero and now it also does not show up in the dropdown. But on audience level we could replace it witch something like: "no restrictions"

bedroesb:
I added min-height, it is for sure better :)

bedroesb:
Current version:
Screenshot from 2021-11-25 13-42-20

Alternative version:
Screenshot from 2021-11-25 17-26-50

floradanna:
About Martin’s comments:
1&2. This text is very important. It is one of the 2 places where it is explained what a tool assembly is. If having it as normal text makes it more visible, I am in favour. Maybe we could even make it into 3 bullet points as suggested by Dom for the landing page.
3. “Affiliation” could be a familiar word for academics. We should not have any “Affiliations”: “ALL”. In that case there shouldn’t be any affiliation.
4. Having “Audience” is controversial since often it is not clear what the audience is: it often depends on affiliations of the user, involvement in specific projects or organisations or initiatives, on collaborations, on how much resource you need, on the instance, on guest accounts and other exceptions, etc… I am not sure a blurb of text would help.
5. I would say that the flags would eventually link to the country pages.

PR #770, opened 2021-12-02
Updating documentation about adding a new tool or resource

bedroesb:
To make everyone's life easier we recommend only one way of adding new tools and resources.


jburel:
The downside of the googlesheet is that it is not possible to view the tools in a given page until it is added to the file under version controls. This is a problem when checking the changes locally.

Will only supporting for example ``tool_and_resource_list.yml`` being an option?
I personally don't think it is a limiting factor and it is much easier to use/understand than the CSV file or even the googlesheet

bedroesb:
`tool_and_resource_list.yml` was first before anything else, but out of request from the community we made the csv. The yaml file was not giving a good overview of what is already there and we wanted to give an option that does not scare contributors. 
CSV also allows for sorting and filtering so duplication's and irregularities are quickly spotted.

But I agree that manipulating a csv is not fun if you are for example using excel or a plain text editor because automatic manipulations will happen or mistakes will be made when listing things. This is why I would only keep the Google spreadsheet.

I know it is not ideal, and locally you can not see your added tools, but the amount of people that preview locally is very small compared to the contributors that purely work trough the GUI of GitHub. I am sorry you are affected by this decision. If the general need will be to change this, I am happy to do so.

jburel:
Thanks for the clarification. I understand 
I will made the changes required locally to view tools or new tag and not commit then :-)

PR #772, opened 2021-12-07
General improvements

bedroesb:
- [x] Adding external link icon
- [x] dependency update
- [x] Github metadata readout for general site urls
- [x] New topnav mechanics to support dropdowns
Screenshot from 2021-12-07 16-59-55


floradanna (approved):

martin-nc (approved):
I guess the external link icon is a matter of taste. Except Wikipedia, I don't think many major sites on the web use it (Twitter, Amazon, CNN, YouTube, NBC, BBC). I wouldn't use it myself, but just make the link text suggest where the link is going. The UK government website puts a lot of time into usability and they dropped it: https://designnotes.blog.gov.uk/2016/11/28/removing-the-external-link-icon-from-gov-uk/. The Norman Nielsen Group don't use it, either: https://www.nngroup.com/. So I think it's preference rather than convention. For me personally it doesn't add anything (except a bit of clutter, perhaps) but I don't have strong feelings about it.

bedroesb:
That is a good point! lets talk about this in the editorial meeting

PR #776, opened 2021-12-13
Machine actionability page

floradanna:
What makes a file machine-actionable?


bedroesb:
I like the title, this can become a hit on google search :+1: More of these please!

PR #801, opened 2022-01-14
Update no_resources.md

korbinib:
* typos
* DSW highlighting


bedroesb:
Because maybe we should merge that one first before more merge conflicts appear

PR #803, opened 2022-01-14
Proposal for DSW integration

bedroesb:
Similar to the training, the metadata ca
Screenshot from 2022-01-14 15-03-59
n be used to list questions in the data stewardship wizard: 

Using this snippet in the metdata:

```yaml
---
dsw:
  - name: File Naming conventions
    uuid: bc5e3dbf-2923-4025-a49a-f204b01d4018
  - name: How will you do file naming and file organization?
    uuid: 8e886b55-3287-48e7-b353-daf6ab40f7d8
  - name: Are you using a filesystem with files and folders?
    uuid: a12aa967-28a5-4a9b-8df8-f7c533205ea4
---
```

:star: [LIVE PREVIEW](https://bedroesb.github.io/rdmkit/data_organisation.html) :star:


floradanna:
@rwwh please, let us know what you think of this suggestion.

rwwh:
This looks good, I am just asking myself whether this is close enough to the relevant position in the page in case there are multiple sections (like with Human Data). Would it be considered in such cases to position an additional link close to the relevant location on the page? I guess the fact that we're using links to sections in the RDMkit pages directly from DS Wizard indicates that sub-page accuracy is needed in some cases.

bedroesb:
I assume also have this problem  for FAIRCOOKBOOK... https://rdmkit.elixir-europe.org/data_transfer.html#more-information 
Different sections having their own toollist,more information section, ... is asking for trouble though with the current jekyll implementation

rwwh:
I see two other ways, both with their own drawbacks. Maybe someone can think of another trick.

1. Split sections on different topics into different pages. 
2. Insert textual references to these context-dependent resources, exactly in the right place. 

Wild thought: Could the reference be given a # context and from that make the template recognize the header under which it should be inserted?

bedroesb:
Today in the editorial meeting we decided to add the links at the more information section automatically as a first step. In the meantime Martin will have a look how we could bring more information things below sections/within the text.

bedroesb:
thank you @MarekSuchanek for your code snippet in parsing the knowledge model!

_config.yml
```diff
@@ -10,7 +10,7 @@ description: "Find the answers to your Research Data Management questions in Lif
 gtag: G-RXQ55EFTTH
 # Google analytics tag
 
-dsw_deep_link_prefix: https://researchers.ds-wizard.org/knowledge-models/dsw:root:latest/preview?questionUuid=
+dsw_deep_link_prefix: https://converge.ds-wizard.org/knowledge-models/dsw:root-rdmkit:2.3.7/preview?questionUuid=
```
bedroesb, line 13: @MarekSuchanek  this can propably also be `https://converge.ds-wizard.org/knowledge-models/dsw:root-rdmkit:latest/preview?questionUuid=` ?

MarekSuchanek: Yes, it can... with the automatic update of links, the `latest` seems to be the optimal way of linking.

bedroesb:
@rwwh Maybe we/you open an issue about sections on pages and its corresponding information. + subpages? I think this pull request is ready to be merged.

PR #813, opened 2022-01-21
Adding a subtitle at the top of the title

### LINKED ISSUE #774: Suggestion to replace titles for 'Data Life Cycle' pages
Currently, for each 'Data Life Cycle' page, heading/title is written in a form of "-ing" noun e.g. Planning, Preserving, etc. 

This can be difficult to understand especially for website visitors that need screen reader to access page content.

It would be better to rename title, for example: Data Collection, Data Preservation, Data Reuse, etc - and, I think it's fine for titles to be a little different/inconsistent to help make content more clear, e.g. for first step: Plan, page title can be Data Management Planning instead of Data Planning.

Happy to know your thoughts! Thank you.

bedroesb:
Screenshot from 2022-01-21 14-44-04
Screenshot from 2022-01-21 14-43-48
Screenshot from 2022-01-21 14-43-05

These are some examples with the font weight. I think if we make it bold, it is to similar with the subheadings

Also part of the title:

Screenshot from 2022-01-21 14-57-50

This PR will close #774


bedroesb:
or even better: Grey
Screenshot from 2022-01-21 15-00-13

PR #817, opened 2022-01-24
Update table: your_task, your_domain

siirif:
All tags in the table did not show on the web page because I had the RDMkit internal your task & domain links wrong (sensitive_data--> sensitive etc.). I tried to fix them. Hopefully they are now correct. Sorry and thanks!! :)


bedroesb:
woops, good catch! https://rdmkit.elixir-europe.org/website_overview.html can help with the tags!

siirif:
Hey thanks Bert! I didn't know that exists, very helpful indeed! I was only checking the cheat sheet!

Issue #748, opened 2021-11-11
Additional audience: collaborators


korbinib:
For tool assemblies with a specific audience of users from one country and their (international) collaborators, the country flag might be perceived as too narrow. 'ALL' + Countryflag might be too wide. Adding a ' collaborators' tag might help.

bedroesb:
I don't 100% understand how this would visually look like?

Nazeeefa:
I would suggest again: remove Audience and just keep Affiliations. It says in description already, for example “for researchers in Finland/Norway” and that’s sufficient to mention what audience it is intended for. You could replace it with "for researchers and their collaborators in Finland/Norway"

bedroesb:
I am ok with leaving out audience and to make it clear in the description. Or there other opinions?

PR #824, opened 2022-02-02
Taking away the audience concept

bedroesb:
This will close the issue #748 
The concept of audience is giving some confusion. The assemblies are meant as inspiration and for this useful for everyone + with the introduction of national resources, it is clear on the page itself which tools are country specific.

To be discussed!


bedroesb:
@korbinib I don't know if everything is fine on your side concerning this pull request?

korbinib (approved):
looks to me along the lines we discussed.

PR #827, opened 2022-02-04
Update se_national_resources.md

YvonneKallberg:
Initial version, not yet final so should not be visible on RDMkit yet


YvonneKallberg:
Build failed, do not know why, but otherwise we are ready for review and publish

bedroesb:
@YvonneKallberg everything looks fine, so I think it is an error on our side. We recently moved the backbone of RDMkit and because of that I had to add some checks, let me try to recreate the problem and get back to you! Sorry for this inconvenience

lauportell (approved):

bedroesb:
It was a combination from a bug on our side where `related pages` has to be present at a tool (good that you stumbled on this bug!) and we noticed that not all page_id s were correct. A tip to find the correct page ID of a page when listing is this [page](https://rdmkit.elixir-europe.org/website_overview.html). Thanks for your contribution!

PR #828, opened 2022-02-04
Modify national pages template

lauportell:
Added some information in the national pages template to guide the user on how to fill up these pages.


Nazeeefa (approved):

wittigue (approved):

korbinib (approved):

bedroesb:
I wonder if we want to mention the fact that the resources in the metadata, their related pages attribute, is used to let it appear on those corresponding pages as part of the national resources section in the toolstable

bedroesb (approved):

PR #835, opened 2022-02-12
Read next section

### LINKED ISSUE #697: Rename "More information"
**Description**
The paragraph "More information" could be renamed as "Related pages and training" in all pages.

bedroesb:
As follow up to the close #697 issue.  During our first UX session and also when Dom was going over our website, people miss quickly that it is really recommended to read the related pages we have selected for them (especially on the data life cycle pages) .

This PR puts more emphasis on this and makes a separate section called `What to read next?` to the pages. This can off cource be changed.

Screenshot from 2022-02-12 09-53-58

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -7,7 +7,7 @@ topnav_title:
 description: "Find the answers to your Research Data Management questions in Life Sciences."
 # Metadata description of the website
 
-remote_theme: ELIXIR-Belgium/jekyll-bootstrap-theme@read-next
+remote_theme: ELIXIR-Belgium/jekyll-bootstrap-theme@1.2.0
 
 gtag: G-RXQ55EFTTH
 # Google analytics tag
--- a/_config.yml
+++ b/_config.yml
@@ -7,7 +7,7 @@ topnav_title:
 description: "Find the answers to your Research Data Management questions in Life Sciences."
 # Metadata description of the website
 
-remote_theme: ELIXIR-Belgium/jekyll-bootstrap-theme@1.2.0
+remote_theme: ELIXIR-Belgium/jekyll-bootstrap-theme@1.2.1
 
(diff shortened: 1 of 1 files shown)


martin-nc:
I'm afraid I'm still a fan of 'Related pages'! 'What to read next?' sounds like there's a journey through the site you should take. I don't think that's the case. What if people already know the stuff in the 'What to read next' pages, and it's the content on the current page that they don't know? In that case 'What to read next' makes no sense. How do we know what people should read next? We don't know what they know.

floradanna:
"Related pages" sounds good indeed.
I would use "Related pages" and merge it.

bedroesb:
@martin-nc Do you agree on separating related pages from the more information? I was not sure if I understood that correctly

martin-nc:
@bedroesb Yes, I was just talking about the "What to read next?" title text.

lauportell:
I like the idea of dividing it and I like the 'What to read next?' but it's true that it might give the false idea to the user that there's a journey through the website.. so maybe 'Related pages' works better just in case.

Nazeeefa:
I am happy with all of it being displayed under 'More information' but I also appreciate the idea of 'Related Pages' as title - to avoid repeating 'RDMkit' in subtitles, you can add it to main title 'Related RDMkit pages', and then subtitles would be e.g. *looking at the screenshot shared above* Your Tasks, and Tool Assembly.

korbinib:
How about: "You might be also interested in..."?

bedroesb:
Screenshot from 2022-02-18 14-42-14
Screenshot from 2022-02-18 14-42-04
Screenshot from 2022-02-18 14-41-01


This PR also contains updates to the labeling of the elixir resources like we discussed in the latest editorial meeting

Issue #754, opened 2021-11-16
Reorganizing the landing page


bedroesb:
With our UX/UI expert looking at the website it was more than ever clear we have to get rid of the block of text on the landing page + people need to be redirected to the tiles with an obvious button

[UX input](https://docs.google.com/document/d/1voKjx6trkDLwvFDlBnH9u07EJyzpXyqbVbaM5cygrmY/edit#heading=h.fg0ux0hduedh)

bedroesb:
Proposal by @floradanna :


RDMkit - The Research Data Management toolkit for Life Sciences
- Data management best practices and guidelines applicable to research projects from the beginning to the end.
- Information, pointers and examples to make your data FAIR — Findable, Accessible, Interoperable and Reusable – by design.
- Tools, resources and training materials used by life sciences research communities for your RDM.

martin-nc:
I'd even go shorter and put Flora's text as the first part of the About page. Maybe something like: "The Research Data Management kit (RDMkit) for LIfe Sciences helps you follow data management best practices and make your data FAIR: Findable, Accessible, Interoperable and Reusable." Then maybe a link to the About page to find out more. 

Here is a comparison of (L to R) how the home page is now, then with Flora's text, then with the shorter text above. Notice how the content below is more visible in the last one, and people can interact with it quicker (e.g. click on the life cycle wheel).

<img width="1000" alt="rdm-screenshots" src="https://user-images.githubusercontent.com/15261745/146519168-c9ee9cd7-c675-4d76-ab7e-1ccc06e81fd5.png">

floradanna:
- I looked at the suggestions given by the UX expert at BH2021 and I made a proposal (see below).
  - the entire tile (data life cycle, your tasks etc) could be clickable and could correspond to the first page of each section (for instance, Plan for data life cycle, Compliance and monitoring for Your tasks etc), at least until we make a page for each section.
- I also realised that we probably don't need a landing page, but we can just have a nice home page that goes directly to the point.
- I also noticed some inconsistencies that could be easily harmonised (I think):
  - Home button in header --> Home button in side bar _is missing_ --> _Page title is missing_
  - About button in header --> About button in side bar (_ok_) --> Page title could be "About RDMkit" (since the titles seem to be different from the side bar names elsewhere)
  - Contribute button in header --> Contribute button in side bar _is called differently! It should be "Contribute" for consistency_ --> Page title _ok_
[...]

PR #837, opened 2022-02-14
New consistent top navigation

bedroesb:
The top navigation in combination with our side menu is currently quite confusingly implemented and not consistent.

- the top navigation acts sometime to switch to sections with other side menus, sometimes not
- the main section has the landing page not listed in the sidemenu, the other (contribute,about) does
- when you are within the contribute section or the main section, you can not keep track visually on the topnav that you are in that specific section of the website

This fixes these problems by letting the topnav act as a REAL navigator of the section main - contribute - about 

You can preview [here](https://bedroesb.github.io/rdmkit/) how this works

This is linked to discussions in #754


martin-nc:
I think this is much better now. There were just a couple of thoughts:

1. I wasn't sure about "Main" because it doesn't say what it contains. Would "Data management" or "Manage your data" work better?
2. There isn't an obvious connection between the main and side menu on desktop. On mobile you have the main menu then the submenu underneath it, with a title like "About menu". This makes the context of the submenu clear. On desktop there is no title on the side menu, so you have to be aware of the main menu item to understand the context. I'm not sure if we can expect people to have this awareness. Maybe the side menu should have a title over it even on desktop widths?  I think it's clearer once you're in a side menu item because the magenta colour links the side menu item with the main menu item. So I don't know how important this point is!

Screenshot-2022-02-15-at-13 59 45

Screenshot-2022-02-15-at-14 00 04

bedroesb:
1. I agree and this is easy to fix
2. I was wondering the same.... But I also didn't know if repeating the topnav section in the sidenav every time is needed. Or we call it Overview or something generic for the landing pages of Main- About - contribute?

bedroesb:
@martin-nc I made some changes as followup on your comments: https://bedroesb.github.io/rdmkit/index.html

Shine19958:
U

bedroesb:
@elixir-europe/rdmkit-editors Hello editors! What do you think about these proposed changed? Should we do it using an overview sidenav item or without? To be discussed tomorrow in the editorial team meeting.

(this message is also used to test out the ELIXIR-europe team RDMkit editors)

korbinib:
I feel this works nicely and declutters the top menu a bit.

bedroesb:
SO the biggest change is that the topnav now acts as a section switcher (each with its own sidebar), and it keeps keeping track if you are with that section (for example: about) when switching pages

bedroesb:
# 1
Screenshot from 2022-02-21 10-37-49
# 2
Screenshot from 2022-02-18 23-58-31
# 3
Screenshot from 2022-02-18 18-34-31
# 4
Screenshot from 2022-02-15 17-42-52

Some try outs

nsoranzo:
Looking at the examples, I'd also add the top level menu entry to the breadcrumbs, e.g. `Data management > Your domain` in the first example. Both should also be links.

bedroesb:
# 5 (from @martin-nc martin)
image

martin-nc:
Or just a dark button (is this 6?!).

dark-button

bedroesb:
@elixir-europe/rdmkit-editors I made a live example of one of the proposals https://bedroesb.github.io/rdmkit/index.html.
Would love to here your feedback! if things don't appear like in the picture, please clear your cache in the browser :) 
Screenshot from 2022-02-23 17-12-37

This PR will also introduce the concept of a menu title + menu title url that is declared in the sidebar yaml file

martin-nc:
@bedroesb Looks good to me. Does the top link on the side menu need to have the extra space below it? I'm not sure if it does. Also, I'd align the page titles with the top of the top menu item, to keep a grid structure and make the page feel better organised. Again, just my take and feel free to ignore!

alignment

bedroesb:
@martin-nc I noticed this too but can I do this without an arbitrary negative margin? 

image

martin-nc:
Well, you can reduce the `line-height` property (which is creating the space). The only trouble is, this will make wrapped heading lines squish together. To counteract this, you could try the `::first-line` pseudo-element, like:

 `h1::first-line {
      line-height: 1; /* or less */
   }`

But I think it's simpler to go with the negative margin, and I guess just in a media query for bigger screens so it doesn't mess up the mobile layout!

lauportell:
I like the most the proposal from the live version @bedroesb 
My only concern now is that, as a user, I would expect that when you click to the RDMkit button you'd go to a landing page and that Data Management would be a only section of the webpage.. Could we do something about this?

bedroesb:
@lauportell , yes, having an overarching landing page that is on top of the current one, the about one and the contribute one. Something we can discuss in the next meeting ;) I think I will implement the feedback of Martin and than merge so we can give feedback by next week

floradanna:
> My only concern now is that, as a user, I would expect that when you click to the RDMkit button you'd go to a landing page and that Data Management would be a only section of the webpage.. Could we do something about this?

Indeed. We should start thinking about this real landing page for the entire website. Also, we should place "Whats new?", "We welcome new contributors!" and "RDMkit in numbers" elsewhere, since they should not be part of the DM section only. Maybe these could be part of the real landing page or something.

bedroesb:
Only problem is, we add 1 click extra to basicly do anything on the website, because that page will not have a sidebar. Will we than add a Home button too on the top nav? From a structural point of view it makes absolute sense, but I don't know for usability...

Nazeeefa:
and then in the home page you will also have to mention what pages and sections the website covers, etc. It might come across a bit repetitive and yes, would add more work for user to navigate website but will also be useful in a way. One solution could be to make 'About' a homepage maybe move "Whats new?", "We welcome new contributors!" and "RDMkit in numbers" to 'About'. Not sure if this is the best solution, worth trying maybe.

Nazeeefa:
and then in the home page you will also have to mention what pages and sections the website covers, etc. It might come across a bit repetitive and yes, would add more work for user to navigate website but will also be useful in a way. 

One solution could be to make 'About' a homepage maybe move "Whats new?", "We welcome new contributors!" and "RDMkit in numbers" to 'About'. Not sure if this is the best solution, worth trying maybe.

Or create a 'Home' page and move some 'must-know' content that, we want reader to be familiar with in first place, from 'About' to 'Home' page. For accessibility, we will have to make sure that home page doesn't include too many links and links that are included are clearly defined (e.g. avoid 'Click here to read more' and have 'Learn more about xyz') if we were to implement this.

Overall, the idea should be to keep webpages organised in a way that allows user to move to different pages/sections easily and that user has a time they need to interact with our RDMkit pages - since there is a lot of information.

This resource might be helpful to learn about web content accessibility https://www.w3.org/WAI/standards-guidelines/wcag/

floradanna:
We could also live without a landing page and keep the DM section as Home. But we would still need to move "Whats new?", "We welcome new contributors!" and "RDMkit in numbers" elsewhere. Actually, something similar to "Whats new?" is already in About (News, Events). "We welcome new contributors!" could fit in the Contribute page, as well as "RDMkit in numbers".

PR #831, opened 2022-02-09
Add structural bioinformatics

gtauriello:
Added page for Structural Bioinformatics domain as discussed in email exchange with Munazah Andrabi (@smza ).

Updated table with tool and resource list at https://docs.google.com/spreadsheets/d/16RESor_qQ_ygI0lQYHR23kbZJUobOWZUbOwhJbLptDE/edit#gid=268211668 using "struct bioinfo" as page ID. From the existing entries, the text of ModelArchive (maintained by my team) was updated as well and new entries were added at the end of the table.

PR #848, opened 2022-02-21
Added contributor details for struct bioinfo

gtauriello:
Following up from #831 : added some metadata for the contributors to the structural bioinformatics page.


gtauriello:
One more thing I had observed when editing the _data/CONTRIBUTORS.yaml file (independent to my additions): some names seem to have badly named fields like "orchid", "Orcid", "Email". I didn't check this exhaustively, but it might be worth adding some automated checks for that...

bedroesb:
Good point about the metadata check. This is good advice and I think we also need it on the metadata in the markdown. I will open an issue for this!

bedroesb (approved):

PR #852, opened 2022-02-24
Gray instead of dark background for menu title

bedroesb:
If I visit the pages now, I don't know if I like the dark background of the menu title. It feels a bit harsh. 

What would you think of :

Screenshot from 2022-02-24 16-51-00

or 

Screenshot from 2022-02-24 16-50-33

Or is this only me?


floradanna:
I would still prefer [this proposal](https://github.com/elixir-europe/rdmkit/pull/837#issuecomment-1046745583), where the section title on the left: i)is very distinct from the buttons below (the text could be entered maybe), ii)has something pink (the line) as the "twin" button on the header and iii)stays the same independently from whatever get clicked after (it doesn't turn dark or anything).

bedroesb:
fact is that we already use the dark header for cards in related pages, training and elixir resources too. It only became part of the menu structure since #837 . We are indeed probably overthinking this :)

bedroesb:
@martin-nc @floradanna THe line version really grew to me, I implemented this on the [infectious disease toolkit](https://www.infectious-diseases-toolkit.org/contact.html) and I like it a lot more than our current implementation. It is less havy for the eye and yes more serving its purpose.

bedroesb:
it would than look something like this:
Screenshot from 2022-03-09 14-57-31

or

Screenshot from 2022-03-09 14-57-13

martin-nc:
@bedroesb Looks good to me, and less heavy as you say. I might suggest the menu title is the same distance from the menu items as the menu items are from each other, so there's a regular pattern of space down the side menu. Or even if it was exactly twice the space it might work better. It's just to increase the sense of pattern or order. Could be just me, though!

bedroesb:
@martin-nc good point about the distance, I took now the same distance as there is between the generals sections and the all tools and resources button.

bedroesb:
Screenshot from 2022-03-09 15-35-20

bedroesb:
This is now using double the width where the spacing between all tools and resources and the rest is triple

bedroesb:
Screenshot from 2022-03-09 15-43-27

This with same font size

bedroesb:
Great :smile:

PR #865, opened 2022-03-07
News in top navigation 

bedroesb:
This is after our discussion from last editorial meeting where it was proposed to have news in the top navigation. I did not make it **News & events** because the landing page if you click on it is only news... + the navigational starts to be cluttered imo.

I am also not sure if splitting events in two different pages (past and upcoming) makes a lot of sense. If there is no event upcoming, this page will be empty. 

[You can preview the website here](https://bedroesb.github.io/rdmkit/)

Screenshot from 2022-03-07 16-42-12


bedroesb:
@martin-nc you were not there at the editorial meeting, what do you think?

bedroesb:
+ what do we do with outreach that contains presentations and videos?

Nazeeefa:
What about presentations and videos? I didn't really understand that bit, sorry.

There are promotion videos for RDMkit users/contributors, maybe a better place for them is Media kit https://bedroesb.github.io/rdmkit/media_kit.html

bedroesb:
@Nazeeefa I indeed wanted to open the discussion if it is clear that this can be found under outreach and if this should be under "news"

bedroesb:
I kinda have to agree with Martin here. Since the existence of the news section on I had to request people to add stuff to it, with me most of the time doing it after the facts .... I don't want to imagine how it will look like after 2 years.

bedroesb:
@Nazeeefa I don't know if there are hard feelings involved in this? :) Otherwise I feel like closing this instead of increasing the granularity for the sake of it :)

Nazeeefa:
Sure :)

PR #868, opened 2022-03-08
Enabling toc on all pages that have more than 2 <H2> headings

### LINKED ISSUE #859: Consistency for ToC
**Description**
Although most of the pages have a "Table of content", few pages don't. Is there a reason or a rational for this? If yes, is it explained somewhere?
For instance, "About" page has a ToC, while "Contribute" doesn't not. Within "Contribute" section, most pages have ToC except "Copyrights guidelines" and "Git way (advanced)".

bedroesb:
This will close #859


floradanna:
This is not applied to the Data Management section though, correct? So, the landing page for Data Management section will look different from "About" and "Contribute" and "News". It is not a problem, just to clarify.

bedroesb:
Good point. The question here is, do we want a TOC in our MAIN website. I thought not ...

bedroesb:
or de we decide that landing pages do not have a TOC. All of them

bedroesb:
@flora, I added `toc: flase` to all landing pages of the top nav sections, for consistency, and removed them everywhere else, what do you think?

bedroesb:
or do we make an exception for only the current landing page?

floradanna:
Not sure what's best. I find ToC useful to have immediately an idea of how much content there is in the page and how much I have to scroll down to read everything. Without ToC I miss some paragraphs because I don't scroll down enough.

So, my personal suggestion would be to have ToC where we have more than 2 headings (including About and Contribute landing pages). An exception would be the Data Management landing page, of course.

bedroesb:
I agree

PR #877, opened 2022-03-15
Sidebar title line

bedroesb:
Currently the sidebar title is living in a rounded box which makes the line look like this:
Screenshot from 2022-03-15 12-05-54
with the hover state like this:

Screenshot from 2022-03-15 12-06-21

@martin-nc gave a suggestion to make this straight:

Screenshot from 2022-03-15 12-06-51

which makes the hover state look like this

Screenshot from 2022-03-15 12-06-57

### DIFF
--- a/_sass/_custom_classes.scss
+++ b/_sass/_custom_classes.scss
@@ -35,6 +35,8 @@ li ul .sidebar_rdm_sub {
     border-bottom: 0.2rem solid $primary;
     font-weight: bold;
     padding-bottom: $input-btn-padding-y - 0.2rem;
+    border-width: 0 0 0.2rem;
+    border-radius: 0;
 }
 
 #side-nav > ul {


bedroesb:
I kinda like that all hover states look the same, but the non hover state looks clean with the straight line though ...

martin-nc:
@bedroesb sorry, I'd forgotten it changes to a button in the hover state! I've complicated things here. The options are:
1. Leave it as it is.
2. Reset the border-radius for the hover condition so it looks like the other buttons on hover.
3. Use a box-shadow to generate the underline. This won't be mitred like a bottom border, but it might be strange if it changes from a straight end to a curved end on hover:
```
.sidebar-title {
    /* border-bottom: 0.2rem solid #c23669; */
    font-weight: bold;
    padding-bottom: .175rem;
    box-shadow: 0 0.2rem 0 0 #c23669;
    border-radius: 0;
}

.sidebar-title:hover  {
    ...
    border-radius: 5px;
    box-shadow: none;
}
```

FWIW I'm not keen on the square box on hover as shown above, and leaving as it is would be better.

floradanna:
I would also leave it as it is.

bedroesb:
ok lets go for the "easy" solution and keep it as it is. Now the title is on a sled ;)

PR #884, opened 2022-03-23
Adding presentations

bedroesb:
This adds the 1 and 3 slides about RDMkit + the presentation given at NIH Data Sharing and Reuse Seminar 2022.

I have a general question for @elixir-europe/rdmkit-editors , I am not sure about the difference between the page Outreach and the page Media kit. Mediakit seems to contain stuff that can be reused right away, while outreach contains more info about what we have done for PR in general. Is this the truth + is this clear for the user?

### DIFF
--- a/pages/about/media_kit.md
+++ b/pages/about/media_kit.md
@@ -2,7 +2,7 @@
 title: Media kit
 ---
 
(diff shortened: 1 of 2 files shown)


bedroesb:
No comments are a good thing I guess, that means it is not total nonsense I tell :). I just clarified this with a few lines on those pages.

Nazeeefa:
Yes that makes sense. Thank you, @bedroesb!

PR #885, opened 2022-03-28
FAIR Cookbook link proposal

bedroesb:
- [x] Work with urls on RDMkit that do not contain `.html` -> more modern, the .html ones are still supported off course
- [x] Generate mapping yml file (one to many) that contains the current links [mapping file](https://github.com/elixir-europe/rdmkit/blob/FCB-proposal/_data/faircookbook_rdmkt_mapping.yml)
    - [x] move to FAIR Cookbook identifiers instead of full links
- [x] Add script to update RDMkit page files bassed on the mapping file, in a similar fashion as DSW
- [x] Update GitHub Action scripts to update the pages weekly together with DSW
- [x] Updating documentation


bedroesb:
This one is approved in the joint editorial meeting to give it a test run, so lets give it a try

PR #887, opened 2022-03-30
Small improvements to the theme

bedroesb:
- files that start with `TEMPLATE_` will by default have following attributes that are declared in the _config file (not visible to the contributor
  ```yaml
      sitemap: false
      search_exclude: true
      no_robots: true
   ```
   This makes sure the templates do not show up in the search of the website and are not indexed by google search
- better table of contents rendering. No visible jump in the content anymore when the threshold of headings is not reached
- `sitemap.xml` is now rendered using a [Jekyll gem](https://github.com/jekyll/jekyll-sitemap), which takes away another worry from the theme
- contributor tiles at the end of the page have correct margins + are smaller
- better text wrapping of resource table headings
- icons for regsitries in the tools and resources table


bedroesb:
Screenshot from 2022-03-30 17-09-41
Screenshot from 2022-03-30 17-09-33

PR #890, opened 2022-04-08
Update no_resources.md

korbinib:


korbinib:
@Nazeeefa @bianchini88  - added some more support services. Maybe we can also add some more detailed description in the same go.

lauportell (approved):

PR #891, opened 2022-04-09
Update used KM for DSW integration

MarekSuchanek:
After releasing `dsw:root:2.4.0` (that includes a set of changes made during the DSW-RDMkit hackathons as well as changes for Horizon Europe DMPs), we decided with @rwwh to keep updating directly `dsw:root` KM instead of having the *Common DSW Knowledge Model (RDMkit WIP)* (`dsw:root-rdmkit`).

### DIFF
--- a/var/dsw_integration.py
+++ b/var/dsw_integration.py
@@ -13,7 +13,8 @@ import yaml
 rootdir = 'pages/'
 DSWQuestion = collections.namedtuple('DSWQuestion', 'uuid, text')
 RDMKIT_PREFIX = 'https://rdmkit.elixir-europe.org'
-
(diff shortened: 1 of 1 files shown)


bedroesb (approved):
that is great! Thanks for pushing this update.

MarekSuchanek:
It seems that the links are still using root-rdmkit 😢

bedroesb:
damn you are right! this bastard: https://github.com/elixir-europe/rdmkit/blob/master/_config.yml#L16 has to be changed too, I totally forgot.

MarekSuchanek:
Should I create a new PR with that change?

bedroesb:
I will make it and tag you in it

PR #900, opened 2022-04-22
Adding support for more training registries

bedroesb:
Logo for zenodo, youtube and the carpentries:
Screenshot from 2022-04-22 18-44-26

more info on the options can be fount in this [PR of the theme](https://github.com/ELIXIR-Belgium/elixir-toolkit-theme/pull/68)

Also registry_url is not needed anymore, since it had not really a purpose


bedroesb:
I guess this is a small change that don't need reviewing? :) Please comment here if you see something weird!

PR #918, opened 2022-05-05
Nicer contributor popovers

bedroesb:
This PR will bring the same style and information from the contributor tiles in the overview page to the popovers at the bottom of each page:
Screenshot from 2022-05-05 16-08-00

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -4,7 +4,7 @@ title: RDMkit
 description: "Find the answers to your Research Data Management questions in Life Sciences."
 # Metadata description of the website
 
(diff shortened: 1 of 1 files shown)


martin-nc (approved):
Looks good to me! I see the popover footer is a different colour in the Contributors' page but I think that's fine. You can't use the grey against the grey background of a "normal" page and the dark blue would be too much in the Contributors' page!

bedroesb:
@martin-nc I took now the same drop shadow as the search dropdown, but yes you are right

bedroesb:
I guess for now I leave it like this. it is an improvement. If it bothers me I make a new PR :)

PR #926, opened 2022-05-12
Removing duplicate tools in tools table 

bedroesb:
Some tools are described both in the main tools table and on a national resource page. This many times with other `page tags` and sometimes even another description/ slightly different name. Duplicates that differ from each other will only cause confusion. 

@elixir-europe/rdmkit-editors I tried to untangle the duplicate tools that got into RDMkit, but please check if I made the right decisions with following rules in mind:
- If a tool is already in the [main tools table](https://docs.google.com/spreadsheets/d/16RESor_qQ_ygI0lQYHR23kbZJUobOWZUbOwhJbLptDE/edit#gid=268211668) and is is accassible for everyone, and devoted to a general audience, the one on the national resource page will be deleted.
- When a tool has a `how_to_access` statement in the national resource page, this one will be kept and the one in the main tool table will be deleted, since it does not have a general audience

I will also add this to the national resource template

General message: I know we gave freedom to the national resource page, but since those tools also end op in the general table, we have to be more stringent. In the future we can always see if we want to add affiliations or how we can let general tools appear on the national resources page.


bedroesb:
This all started in : https://github.com/elixir-europe/rdmkit/pull/925

korbinib (approved):

bedroesb:
@elixir-europe/rdmkit-editors I made some changes to the theme in how to handle these national tools, you can use this as preview: https://bedroesb.github.io/rdmkit/be_resources

The syntax for the metadata in national pages looks now like this, it has the `national_resources` object and the `main_resources` object if needed. `main_resources` are just a list of tools from the main resources table and is used to refer to other tools, `national_resources` is for the national tools like we used them before.

```yml
national_resources:
  - name: GFBio Data Management Plan Tool
    related_pages:
      your_tasks: [DMP]
    url: https://www.gfbio.org/plan

main_resources:
  - e!DAL-PGP
  - FAIRDOM-SEEK
  - PANGAEA
```
national resources now also support registries, which can be used like this:
```yml
national_resources:
  - name: Resource name
    description: A general description about the resource
    how_to_access: explantation on how you can access this resource
    instance_of: GitHub
    related_pages:
      example_pages: [gp3, gp1, gp2]
    registry:
      biotools: bioconda
      tess: Bioconda
```

bedroesb:
To do: agree on the texts/syntax + update the template so it is clear

lauportell (approved):

floradanna:
Instead of `main_resources` we could name it `ref_main_tool_and_resource` ?

smza:
> @elixir-europe/rdmkit-editors I made some changes to the theme in how to handle these national tools, you can use this as preview: https://bedroesb.github.io/rdmkit/be_resources
> 
> The syntax for the metadata in national pages looks now like this, it has the `national_resources` object and the `main_resources` object if needed. `main_resources` are just a list of tools from the main resources table and is used to refer to other tools, `national_resources` is for the national tools like we used them before.
> 
> ```yaml
> national_resources:
>   - name: GFBio Data Management Plan Tool
>     related_pages:
>       your_tasks: [DMP]
>     url: https://www.gfbio.org/plan
> 
> main_resources:
>   - e!DAL-PGP
>   - FAIRDOM-SEEK
>   - PANGAEA
> ```
> 
> national resources now also support registries, which can be used like this:
> 
> ```yaml
> national_resources:
>   - name: Resource name
>     description: A general description about the resource
>     how_to_access: explantation on how you can access this resource
>     instance_of: GitHub
>     related_pages:
>       example_pages: [gp3, gp1, gp2]
>     registry:
>       biotools: bioconda
>       tess: Bioconda
> ```

@bedroesb does this main the main_resources will be displayed in the NR page ( i.e. referenced from the main tools table)?

bedroesb:
@smza correct!

bedroesb:
I would be nice to have some feedback on the sentences/titles I chose:

## Tools and resources
### Tailored to Belgium
Developed and/or deployed by Belgium.

[tool table with national tools]

### Belgium is also involved in
These tools and resources can have a lead from Belgium or Belgium is involved in the development.


[tool table with from in the mains tools table]

bianchini88:
In terms of phrasing, I would prefer something like "tailored to users in Belgium" instead of "tailored to Belgium". 

I think the provider of a resource/service (e.g. an ELIXIR node, or a national computing centre) should also be indicated explicitly in the registry.

bedroesb:
@bianchini88 I hear you, I just think that there are already so many attributes already:

```
  - name: Resource name
    description: A general description about the resource
    url: 
    how_to_access: explantation on how you can access this resource
    instance_of: GitHub
    related_pages:
      example_pages: [gp3, gp1, gp2]
    registry:
      biotools: bioconda
      tess: Bioconda
```

bianchini88:
@bedroesb I still think it is useful to clearly highlight which entity is providing a certain service, as users are often very confused about this. This might as well end up in the "description" if we do not want an additional attribute.

bedroesb:
I like the idea of adding it to the description! :)

bianchini88:
We "only" need to make sure people do it!

smza:
> > I think the provider of a resource/service (e.g. an ELIXIR node, or a national computing centre) should also be indicated explicitly in the registry.
> 
> I remember asking exactly this question during the editorial meeting where this was discussed and the answer was that there was no need to further specify institutions and organisations in the metadata. So, I think it can be done in description if it is relevant.
> 
> If we want to change the use case, we need to discuss again and re-design the implementation.

+1 @floradanna

PR #934, opened 2022-05-27
Popover grey + affiliation in firefox fix

bedroesb:
- Making use of latest version theme
- Fix for affiliation tiles where svgs without width are not shown
- Less hard contributor popover footer, making use of the same color as in the contributor panels:
 Screenshot from 2022-05-27 17-34-33


bedroesb:
The latest version of the theme also has theme defaults so not all variables have to be declared

PR #940, opened 2022-06-10
Patch 7

erwanlefloch:
just a small modification on the reference to the recipe


bedroesb:
Merci!

bedroesb:
@erwanlefloch I am also planning to make the link from the tools assembly to this FAIR Cookbook recipe

erwanlefloch:
@bedroesb Nice! Good idea

PR #947, opened 2022-06-15
Getting rid of the country file 

bedroesb:
The assembly overview page was made before the country pages. Back than I had to add a mapping file to keep track of country names and there 2 letter ISO code. With the country pages in place I can read out the title to know the country name. 
This takes away the dependency of the country file + the workflow that updates the country file. 

Small disadvantage, if a country is listed as affiliation that does not yet have a country page, there will not be a country name tooltip for that country. But I guess this can stimulate the creation of country pages?


martin-nc:
I suppose I'd start with the end point: the user experience. It's true a tooltip is not a big deal, but I guess if there is one missing because we don't have a country page, isn't that a fault in the system? It doesn't feel very robust.

It depends on how rapidly these country pages and assemblies are going to grow, and which will come first. 

You can always have a policy that says, "If you want to add a tool assembly, you have to write a county page first"! I don't think that will be popular, though.

bedroesb:
true true ....the flag would off course still be there. It is only the tooltip that will be missing. I just noticed though that the ones that have no country page are disabled: https://rdmkit.elixir-europe.org/tool_assembly and the tooltip is not working

bedroesb:
Lets close this, normally I improved the lookup for country names so tools can still get updated if the API of for  country iso codes is down (which  happened already)

PR #983, opened 2022-09-20
Remove ambiguity from the "Titles" entry

martin-nc:
"Titles" and "Headings" could be confused, so I've added extra text to distinguish them. This was suggested for the IDTK: https://github.com/elixir-europe/infectious-diseases-toolkit/issues/47.


bedroesb (approved):
This is clearer indeed

bedroesb:
I will also apply this to IDTk :)

PR #984, opened 2022-09-22
Update fi_resources.md

siirif:
Changed the Icebear description to the Nazeefa's suggestion!

### DIFF
--- a/pages/national_resources/fi_resources.md
+++ b/pages/national_resources/fi_resources.md
@@ -89,7 +89,7 @@ national_resources:
       your_tasks: [data analysis]
     url: https://research.csc.fi/computing#cloud-computing 
   - name: IceBear
-    description: A RDM tool for structural biology. It covers crystallization trials, crystal fishing and shipping, as well as linking crystals, datasets, and PDB depositions for full chain of custody. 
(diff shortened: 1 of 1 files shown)


bedroesb:
Hey Siiri, The icebar tool you added in the main tools table, that is the national tool that is accessible/targeted for all countries right?

If so it can be taken out on the national resource page since we don't duplicate tools.

siirif:
Hi Bert! Ah sorry! I was thinking it would be good advertisement also on the FIN page as there are many FIN collaborators. However, I'm not sure whether it is open for "everyone", but I checked that to use it one can apply Instruct-ERIC access: https://instruct-eric.eu/platform/macromolecular-crystallisation-oulu-finland/) and there are at least 13 member countries in Europe + EMBL.

bedroesb:
@lauportell @Nazeeefa what do you think, do we keep the tool here or in the main tools table?

Nazeeefa:
It's fine to keep it. I see HPC and Fingenious in the list, and those services are for researchers outside Finland too.

bedroesb:
keep it where? :)

Nazeeefa:
National resources page for Finland :D

bedroesb:
@Nazeeefa if you think this is ok, you can approve and merge ;)

lauportell (approved):

PR #991, opened 2022-09-25
add link to plant phenomics assembly

erlefloch:
I just added a link in the related pages section


bedroesb:
This is not allowed sadly since we have strict rules on which links can be linked to which type of page: https://rdmkit.elixir-europe.org/editorial_board_guide#related-pages

I will have to check the existing ones on the "plant phenomics assembly" one

erlefloch:
oh ok, alright. Then we should remove the link to the plant genomics assembly on the plant phenomics page indeed.

PR #993, opened 2022-09-26
Update plant_phenomics_assembly.md

erlefloch:
removal of:
- dropdowns lists of links not displaying correctly
- the link to geno toolkit (not allowed unfortunately)

### DIFF
--- a/pages/tool_assembly/plant_phenomics_assembly.md
+++ b/pages/tool_assembly/plant_phenomics_assembly.md
@@ -8,9 +8,18 @@ related_pages:
   your_tasks: [metadata, data publication]
   your_domain: [plants]
 training:
-  - name: Training in TeSS
+  - name: MIAPPE training in TeSS
     registry: TeSS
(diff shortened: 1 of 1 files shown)


erlefloch:
@bedroesb these are the updates that we just discussed

korbinib (approved):

bedroesb:
I agree it is unfortunate for the link towards the other Assembly/. I can always bring it up in the editorial meeting. Thanks!

bedroesb (approved):

bedroesb:
@erlefloch I see you took them fully out, but mabe we can refurbish the content in an other format ... @korbinib Do you think we should keep the content of the dropdowns?

korbinib:
Maybe they could be split between tools, references and training material?

erlefloch:
> Maybe they could be split between tools, references and training material?

good idea !

erlefloch:
@korbinib @bedroesb I have splitted the content.

bedroesb:
Let's merge.

PR #994, opened 2022-09-26
Update tools plant pheno assembly

bedroesb:
@erlefloch so you know the tools will soon be visible on the page ;)


erlefloch:
@bedroesb thanks!
btw I saw you changed the page_id, this can be changed on the plant sciences domain page too (currently it is linking to `tool_assembly: [plant geno assembly, plant_pheno_assembly]`)

bedroesb:
good catch, I'm on it

PR #1019, opened 2022-10-20
Adding testimonials to the RDMkit

bedroesb:
I would like to use this pull request to try out some things.

My first idea was to try out to display them on the home page (with or without carousel) and link from there to an actual testimonial page. You can see how this would look like/follow the progress here: https://bedroesb.github.io/rdmkit/

Current preview:
Screenshot from 2022-10-20 14-34-17


Nazeeefa:
Looks great, @bedroesb 
Just wondering if links/hyperlinks with buttons can also be added? either below or inside the text area of a testimonial - thinking here about links to external websites (e.g. for articles, publications, etc)

bedroesb:
@Nazeeefa that should not be a problem!

bedroesb:
I am not a fan of the word 'impact', gives me funding-language-vibes :) but I do like one, or a variation of one!

martin-nc:
I would focus on "you" and "people" rather than "we" or "users". I think we need to make it clear this section isn't for us or for funders, but it's to help people get the most out of the site. Maybe:
7. How people use RDMkit
8. How you can use RDMkit (tagline: "Get ideas from people who already use the site."
9. Ways you can use RDMkit
10. Ways people use RDMkit
11. How others use RDMkit

Nazeeefa:
@bedroesb Ah ok, I see that. I was thinking of how RDMkit helped/inspired, when using the term 'impact'.

Thanks @martin-nc :)
> I would focus on "you" and "people" rather than "we" or "users". I think we need to make it clear this section isn't for us or for funders, but it's to help people get the most out of the site.

I thought this could be done as a part of description, while we keep title concise and clear

PR #1028, opened 2022-10-28
Add hackathon as event

bedroesb:
Screenshot from 2022-10-28 18-07-53


floradanna:
I would used a different message. More like " Take the opportunity to revise RDMkit pages; editors will be available in two time slots to help you with it."

Nazeeefa:
Yes @floradanna +1

PR #1041, opened 2022-11-07
Update sensitive_data.md

smza:
Changed the name from Training in TeSS to RDMbites


pages/your_tasks/sensitive_data.md
```diff
@@ -6,7 +6,7 @@ page_id: sensitive
 related_pages: 
   tool_assembly: [TSD, Covid-19, transmed]
 training:
-  - name: Training in TeSS
+  - name: RDMbites
```
bedroesb, line 9: This is not correct I think? We link here to a general query

PR #1048, opened 2022-11-10
Adding persistent identifiers based on page_id

### LINKED ISSUE #818: DOI for our pages?
In meeting with FAIRcookbook, it is already mentioned that one of the problems with RDMkit is that we don't have a DOI. I would like to use this issue to brainstorm in how we can make our pages citable. 

- FAIRcookbook generates DOIs/URLS using [W3id.org](https://github.com/perma-id/w3id.org), is somebody familiar with this? Can this be fully automated if a new page is made?
- We should put redirects in place if we change the url/filename of our pages
~- Adding last updated (linked to #766 )~

bedroesb:
This is a proof of concept to have persistent identifiers on RDMkit, this will close #818 

In this scenario we use the `redirect_from` attribute on every page to serve the page at `https://rdmkit.elixir-europe.org/pid/page_id` 

On the w3id side we redirect from `https://w3id.org/rdmkit/page_id` to https://rdmkit.elixir-europe.org/pid/page_id, which wil redirect to https://rdmkit.elixir-europe.org/the correct page.

A discussion we will need to have is: what type of page identifiers are we going to use? 
In theory we want numerical ones that do not contain info related to the page, so it is content independent. But this will be difficult to use in "related pages", since one will have to use a mapping table all the time to fill them in.... 

If we keep on using text-identifiers , we have to get rid of our spaces!

### DIFF
--- a/pages/data_life_cycle/TEMPLATE_rdm_cycle.md
+++ b/pages/data_life_cycle/TEMPLATE_rdm_cycle.md
@@ -4,7 +4,7 @@ contributors: [<!---REPLACE THIS with comma separated list of contributors--->]
 search_exclude: true
 description: <!---REPLACE THIS with a one sentence description of the page--->
 page_id: <!---REPLACE THIS with a shortened page name, with small letters and spaces, or an acronym in capital and small letters--->
-redirect_from: /pid/<!---REPLACE THIS with a shortened page name, with small letters and spaces, or an acronym in capital and small letters--->
+redirect_from: /pid/<!---REPLACE THIS with page_id--->
 related_pages: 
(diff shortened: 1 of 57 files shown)


bedroesb:
a live working example: https://w3id.org/people/bedroesb/rdmkit/analyse will hopefully bring you to the analyse page

korbinib:
favouring to get rid of spaces (which is a good idea in the first place imho)

bedroesb:
In previous WP3 meeting we decided to not have 'new' identifiers for RDMkit (so we don't need the redirects added here in this PR) Instead we will make use of the current filenames as page_id. In many cases this will make it a bit longer, but the advantage is that one can know the page_id from the page url, which makes tagging of pages or tools a lot easier.

PR #1050, opened 2022-11-12
Bio.tools and fairsharing metadata try-out

bedroesb:
This PR add automation to RDMkit to pull the description and url of a tool or resource when the bio.tools or FIARsharing ID is given. This using information on bio.tools and FAIRsharing using their respective API enpoints.


bedroesb:
It is decided in the editorial meeting that for now, we do not implement this system. The findings from this PR will be put in a discussion topic

PR #1056, opened 2022-11-15
Fixed "how to contribute" link

fbacall:
and restyled Elixir to ELIXIR

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,7 @@ This project is governed by a [code of conduct] (https://github.com/elixir-europ
 ## How to contribute 
 
 You can create new pages or report errors or typos in two ways:
- - Use the GitHub web interface (see How to contribute)
+ - Use the GitHub web interface ([How to contribute](https://rdmkit.elixir-europe.org/how_to_contribute))
  - Send a text file with your suggested changes to the editors (rdm-editors@elixir-europe.org). 
 
 ### Reporting a bug or a typo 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -18,4 +18,4 @@ Bugs or remarks are tracked as GitHub issues. You can create an issue and choose
 
 ### Adding a new page or changing an existing page
 
-See our [How to contribute page](how_to_contribute).
+See our [How to contribute page](https://rdmkit.elixir-europe.org/how_to_contribute).
```
floradanna, line 21: I think we only use the page name and not the full url, @bedroesb correct?

floradanna:
Should the link to the contribute page in line 12 be added as well?

CONTRIBUTING.md
bedroesb: In theory yes, but I think many people end up on this page in its 'markdown' form on GitHub, and than people might want to be directed to the actual RDMkit page with contribute info. But if they find the page thought RDMkit, than the old linking is better

fbacall: I used the full URL because if you visit https://github.com/elixir-europe/rdmkit/blob/master/CONTRIBUTING.md from GitHub the relative link doesn't work.

Is `CONTRIBUTING.md` also served by the RDM-Kit website, or only for viewing in GitHub? If it's the former, I'm not sure what the best solution is here.

floradanna: oh, yes. sorry. my mistake.

floradanna (approved):

PR #1060, opened 2022-11-18
Add supported registries for training attribute

bedroesb:
The supported registries that can be used in the `registry` attribute are: *YouTube*, *Zenodo*, *Carpentries* and *TeSS*.
  
  and linking out to the metadata manual in the metadata section itself


bedroesb:
Since this is obvious extra documentation, I dare to merge this one without approval :)

PR #1075, opened 2022-12-05
Update human_data.md

greg-r-sib:


bedroesb (approved):

bedroesb:
Although we have to discuss further on having countries in a domain page, since we were doing it on the moment you proposed this change, I will go ahead and merge this. But this might change in the future! Thanks for your contribution!

bedroesb (approved):

PR #1107, opened 2022-12-08
Updating TSD Tool Assembly page

Nazeeefa:
Replaced links for ELIXIR AAI to Life Science AAI, and updated "Data Sharing and Preservation" section.


bianchini88 (commented):
Do we want to use Life Science AAI or Life Science Login? I think the latter is more common, although they both are in use. The shortening to LS Login is also very common.
As a link, we could use this one https://lifescience-ri.eu/ls-login/ instead of the ELIXIR page.

Also, is NFEGA still an accepted acronym or should we just be using FEGA Norway?

korbinib:
We should stick with LS|Life Science Login (changed accordingly).
I fell NFEGA has internal slang, removed it for now.

Nazeeefa (commented):
Replaced FEIDE with Feide

PR #1121, opened 2022-12-08
Use GitHub Actions for Jekyll and GitHub Pages

MarekSuchanek:
As a followup to [discussion](https://github.com/elixir-europe/rdmkit/discussions/1064), this PR brings GitHub Actions that build the site with Jekyll and then publish it (on a push to the default branch) or via workflow dispatch (manually on any branch).

It simplified development as it can be easily run/developed locally with use of the supplied Makefile. Also, in the `development` env, there is a bar at the top showing actual repo and commit which can be handy while having multiple versions opened.

To allow GitHub Actions to be used for building GitHub Pages (in a fork):

1. Go to *Settings* > *Pages* and enable *GitHub Actions* as a source
2. Go to *Environments* > *github-pages* and remove the rule under *Deployment branches* if you want to deploy other branches than master via Workflow Dispatch (manually triggered action)
3. Enable GitHub Actions (and use Workflow Dispatch)

Finally, I made some changes in the workflow files to make them more consistent and readable. I do not like names of the other two workflows, but that is something that we may discuss and resolve later...


bedroesb:
@MarekSuchanek long time ago! I wonder if you could bring over the ruby script to fix the branch name to this repo: https://github.com/ELIXIR-Belgium/elixir-toolkit-theme-plugins This repo is used for the custom scripts that will be included in a 2.0 version of the theme that is used in RDMkit, that will add better GitHub Actions support. Let me know if you need any help!

MarekSuchanek:
@bedroesb OK, I will try to work on it during this week... But it is a question whether you still want to support just GitHub or add some support for others (to name it, e.g. `GIT_BRANCH` instead of `GITHUB_BRANCH` but also check if there is something else get by the `jekyll-github-metadata` which should be possible to override with envvars)?

bedroesb:
Good question, we use it internally also for GitLab, so git is maybe better. In the dev-info banner you also make use of `site.github.build_revision` so we should make sure that is also the correct one.

MarekSuchanek:
I opened https://github.com/ELIXIR-Belgium/elixir-toolkit-theme-plugins/pull/1

bedroesb:
@MarekSuchanek thanks a lot! I still have to respond to your linked PR, but I guess after that one is merged, we can cleanup this PR. We should add a disclaimer to this PR that it should not be merged until a version is released of the toolkit theme that contains your plugin as dependency.

MarekSuchanek:
@bedroesb Now I cleaned it up here based on the PR in plugins... can you check? By the way, this PR is still as "Draft" so I think it is good as the disclaimer for not merging.

I guess once you release a new version of the theme (2.0) with use of the plugin, we can wrap it up here?

bedroesb:
@MarekSuchanek thanks a lot, indeed!

bedroesb:
@MarekSuchanek I was thinking of merging this as an intermediate step for #1249 . I the Makefile needed? + what is it for ? :)

MarekSuchanek:
> @MarekSuchanek I was thinking of merging this as an intermediate step for #1249 . I the Makefile needed? + what is it for ? :)

It is useful for local development as shortcuts so you do not need to spend time figuring out what env vars to set and what are the commands to run 🤷🏻‍♂️

bedroesb (approved):

bedroesb:
Basically adding you PR notes to the contribute pages

bedroesb:
@MarekSuchanek Thanks a lot!

PR #1125, opened 2022-12-08
Updating figure to include brand blue

martin-nc:
The [data brokering page](https://rdmkit.elixir-europe.org/data_brokering) image was using a slightly different blue.


bedroesb (approved):

bedroesb:
Thanks for updating this!

PR #1151, opened 2022-12-08
Updating the broken Quick suggestion form

martin-nc:
The current form no longer works, so this replaces it.


bedroesb (approved):

bedroesb:
Great! Thanks

PR #1162, opened 2022-12-16
Replacing all etc with etc.

### LINKED ISSUE #1127: Replace 'etc' with 'etc.' on all pages.
Several pages write 'etc' while it should be 'etc.' regardless  of any additional punctuation that may follow.
Could you please search all pages and add '.' if it is missed?

bedroesb:
Sorry for all the spam! 

As mentioned in this issue #1127, etc is always followed by full stop. I replace all occurrences of etc to fix this.

This will close #1127.


martin-nc:
I think we should discourage people from using "etc." since it can be a bit vague and lazy. It assumes people already know what you mean, so you don't have to write it. Maybe people don't know what you mean. Style guide usually warn against it (e.g. https://insidegovuk.blog.gov.uk/2016/07/20/changes-to-the-style-guide-no-more-eg-and-ie-etc/). It's okay for now, but maybe it's worth asking people in the style guide to avoid it if possible?

bedroesb:
@martin-nc I totally agree martin. We could have a look if there are quick fixes in the ones that get changed here.

pages/contribute/style_guide.md
```diff
@@ -20,6 +20,7 @@ In general, we follow the European Commission's [Web Writing Style Guide](https:
   * **Datasets:** not "data sets".
   * **Email:** not "e-mail".
   * **Email addresses:** spell these out and make the email address the link e.g. [rdm-toolkit@elixir-europe.org](mailto:rdm-toolkit@elixir-europe.org). Do not hide the email address behind a word or phrase like "contact us".
+  * **etc** should be avoided. Try using ‘for example’ or ‘such as’ or ‘including’ at the start of a listing.
```
bedroesb, line 23: @martin-nc What do you think of this line?

bedroesb:
@elixir-europe/rdmkit-editors If somebody else can have a look and approve this, this can be merged :)

bianchini88 (commented):
There is no consistent usage of commas with "etc" at the moment, i.e. we both have "A, B, etc." and "A B etc". I suspect the former is the correct one.

pages/contribute/style_guide.md
martin-nc: @bedroesb Sorry - I missed this comment. Yes, that's okay. Just a tiny tweak:

> * **Etc.** should be avoided. Use ‘for example’, ‘such as’ or ‘including’ instead.

martin-nc:
@bianchini88 Yes, I think the former one is right (the one with the commas). I couldn't access the [EC style guide](https://commission.europa.eu/system/files/2022-12/styleguide_english_dgt_en.pdf) to check what it says, because it's become inaccessible! I found it in [another location](https://commission.europa.eu/system/files/2022-12/styleguide_english_dgt_en_2022_11_28.pdf) but the PDF won't load. Hopefully that's just temporary.

Anyway, I guess you'd:
1. Put a comma before it if it was in a list, like "A, B, etc."
2. Put a comma after it, like "They provide A, B, etc., as well as more specific services."

But I think the best thing is to not have etc. at all!

martin-nc (approved):
Looks okay to me. I was itching to replace them, but maybe another day!

PR #1173, opened 2023-01-13
Better way of using GH Actions

bedroesb:
With these changes the weekly registry lookup will not be run on a fork when GH actions are enabled + this is a more robust solution towards the future when we transition towards the use of GH Actions for deployment.


bedroesb:
This way of working is already implemented in the Infectious diseases toolkit

PR #1183, opened 2023-01-25
Update metadata_management.md

bianchini88:
Removing "schemapedia", as the resource does not seem to be maintained and the webpage is no longer up.


bianchini88:
Removed from spreadsheet as well.

floradanna (approved):

bedroesb:
Great :ok_hand:

PR #1184, opened 2023-01-26
Add ETT branding in footer of RDMkit + training fix

### LINKED ISSUE #1172: check metadata for training in these pages
**Describe the bug**
Metadata for training might be wrong in these 2 pages: provenance and dm coordination

**To Reproduce**
Screenshot 2023-01-13 at 13 27 06

bedroesb:
This will close #1172 and bring the mention of the [ELIXIR Toolkit Theme](https://elixir-belgium.github.io/elixir-toolkit-theme) in the footer:

Screenshot from 2023-01-26 10-05-14

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@ _pdf
 .DS_Store
 .idea
 .bundle/
-_site
-/vendor
+_site/
+vendor/
 .vscode/
 Gemfile.lock
--- a/_config.yml
+++ b/_config.yml
@@ -4,7 +4,7 @@ title: RDMkit
 description: "Best practices and guidelines you can use for FAIR management of your research data."
 # Metadata description of the website
 
-remote_theme: ELIXIR-Belgium/elixir-toolkit-theme@new-branding
+remote_theme: ELIXIR-Belgium/elixir-toolkit-theme@1.22.0
 sass:
   style: compressed
   sourcemap: never
--- a/_config.yml
+++ b/_config.yml
@@ -4,8 +4,7 @@ title: RDMkit
 description: "Best practices and guidelines you can use for FAIR management of your research data."
 # Metadata description of the website
 
-remote_theme: ELIXIR-Belgium/elixir-toolkit-theme@1.21.0
-
+remote_theme: ELIXIR-Belgium/elixir-toolkit-theme@new-branding
 sass:
   style: compressed
   sourcemap: never
--- a/_config.yml
+++ b/_config.yml
@@ -4,7 +4,7 @@ title: RDMkit
 description: "Best practices and guidelines you can use for FAIR management of your research data."
 # Metadata description of the website
 
-remote_theme: ELIXIR-Belgium/elixir-toolkit-theme@1.22.0
+remote_theme: ELIXIR-Belgium/elixir-toolkit-theme@1.23.0
 sass:
   style: compressed
   sourcemap: never
--- a/index.html
+++ b/index.html
@@ -8,6 +8,7 @@ layout: none
 {% include head.html %}
 
 <body>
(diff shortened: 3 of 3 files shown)


bedroesb:
I tagged everyone to highlight this change, this was brought up during the hackathon, that people are not aware that there is this layer behind RDMkit that is easy to reuse, to spawn RDMkit-like website with different content/usecases.

martin-nc (approved):
I wonder if "Powered by" is the right phrase. Usually this is for the system that created the site e.g. Wordpress, Joomla, or in this case Jekyll. Maybe "Built with"? 

Perhaps there can be a black and white version for those who want the footer to have a colour?

As for promoting the theme, maybe it could be mentioned in the About page? You could even have some suggested text for the template, in case people are willing to include in it in their About page  - just a note at the bottom of the page.

bedroesb:
@martin-nc concerning the color, problem is a bit that I don't know which color the background is, and on a dark footer white would be better, on a light one, black would be better.... Except if we use an inline svg and I take the same color as text :thinking: 

A mention in the about is I think a good idea! I am ok with build by instead of prowered by......

martin-nc:
With the colour, I was thinking people would get a choice. If the logo comes with the template, though, then I'd be tempted to make it greyscale like the CC logo - just a black/dark grey circle, a white background and the letters in black/dark grey. Then it doesn't matter what colour people use. For me, the current logo stands out too much and I would simply remove it because I wouldn't want it to distract from the things I want to stand out. If it was more discreet, I'd be more inclined to leave it in. That's just me, though. Feel free to ignore!
ett-screenshot

floradanna:
If "Built with" is more correct, I am fine with it.

bedroesb:
I also tried the white letters in a dark ball: 

Screenshot from 2023-01-26 16-27-11

Which I like more on a light background personally! 
Screenshot from 2023-01-26 16-24-34

Nazeeefa:
I agree with mentioning it in About page.
Regarding the ETT logo, it looks good both on light and dark background - the current version of footer has dark background so it should work fine :) 

In terms of its location, I'd suggest to put it in a place where it says "Contributions from the 23 ELIXIR Nodes"* or after "[Institutes, projects and funders](https://rdmkit.elixir-europe.org/support)" so, it is also prominent and easy to find - it would also be helpful to have it linked to GitHub repo of the ETT. You could also maybe mention "(For Developers)" so people who want to use the theme can easily find it, but this can also be described in 'About' page.

*(This can be moved to the text on 'Contributors' page e.g. "This project would not be possible without the many amazing community contributors, _including contributions from 22 ELIXIR Nodes_")

martin-nc:
For the logo colour, maybe the most flexible is to outline the text, make the SVG inline, and so allow people to make it whatever colour suits their theme with the CSS `fill` property. You'd have to use the fill attribute to set the colour in initially, of course (e.g. `<circle fill="#0d6efd" class="cls-1" cx="38.56" cy="38.56" r="38.56"/>`). Then after that people can just do:

`.cls-1 {fill: red /* or whatever */;}`

I'm not sure how many people would use this, or if it's worth it, but I guess this gives people the most options.

bedroesb:
@Nazeeefa thanks for the feedback! Thing is that this notification is meant to appear on all footers of deployments, using the theme. The location I gave it now is not variable and will always be there. If I would put it with the other bullet points it is difficult... Because other deployments might have an image there. 

@martin-nc I will look into the inline svg option!

bedroesb:
@martin-nc I implemented the inline version where the color is set with a variable on the latest version of the theme. I would just want to sort out some things concerning the cookie banner before I make a new release. I was thinking of merging this version to gather some early feedback :thinking:

bedroesb:
Ok I created a new release and am making use of the suggested changes. Lets merge!

PR #1188, opened 2023-02-07
Fixing broken link to the EC style guide

martin-nc:
The link for this guide has changed a couple of times recently. [Yet another EC style guide](https://wikis.ec.europa.eu/download/attachments/6824833/commission_style_guide.pdf?version=1&modificationDate=1594633342434&api=v2) exists, but maybe we're okay with the two we recommend.

### DIFF
--- a/pages/contribute/style_guide.md
+++ b/pages/contribute/style_guide.md
@@ -2,7 +2,7 @@
 title: Style guide
 ---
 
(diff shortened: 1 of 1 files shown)


bedroesb (approved):

bedroesb:
thanks for checking on this!

Issue #729, opened 2021-11-05
Use schema/schemas instead of scheme/schemes


floradanna:
**Link to the page of interest**
all pages

**Describe what you would change to the page**
Use schema/schemas instead of scheme/schemes

bedroesb:
Something to than also add to the style guide?

martin-nc:
> Use schema/schemas instead of scheme/schemes

Yes, I'll add to the style guide.

Issue #1130, opened 2022-12-08
Systematically harmonise "schema(s) or scheme(s)"


floradanna:
**Link to the page of interest**
many pages

**Describe what you would change to the page**
Systematically harmonise "schema(s) or scheme(s)".

@elixir-europe/rdmkit-editors should we vote to decide which one to use?

martin-nc:
They have slightly different meanings. "Scheme" is more of a plan, whilst "schema" can describe structure and data types within the structure e.g. a "database schema" or an "XML schema". I'm guessing it should be "schema".

PR #1194, opened 2023-02-09
Change scheme to schema

bedroesb:
As discussed in #729, the correct word is schema.

### DIFF
--- a/pages/data_life_cycle/preserving.md
+++ b/pages/data_life_cycle/preserving.md
@@ -54,7 +54,7 @@ When preparing data for preservation several requirements need to be fulfilled.
   * Include information about provenance.
   * Include sufficient licensing information.
   * Ensure that data is well organised.
-  * Ensure that a consistent naming scheme is used.
+  * Ensure that a consistent naming schema is used.
   * Use standard, open source, file formats instead of proprietary ones.
 
 If you need to preserve non-digital data (e.g. paper), consider whether digitalising the data is feasible or consult with data management support services in your institution.
--- a/pages/data_life_cycle/preserving.md
+++ b/pages/data_life_cycle/preserving.md
@@ -54,7 +54,7 @@ When preparing data for preservation several requirements need to be fulfilled.
   * Include information about provenance.
   * Include sufficient licensing information.
   * Ensure that data is well organised.
-  * Ensure that a consistent naming schema is used.
+  * Ensure that a consistent naming convention is used.
   * Use standard, open source, file formats instead of proprietary ones.
 
 If you need to preserve non-digital data (e.g. paper), consider whether digitalising the data is feasible or consult with data management support services in your institution.
--- a/pages/your_domain/plant_sciences.md
+++ b/pages/your_domain/plant_sciences.md
@@ -108,7 +108,7 @@ It is recommended that metadata collection is contemplated from the start of the
   * [e!DAL](https://edal.ipk-gatersleben.de/) is a free data management platform for which MIAPPE templates are in development.
   * The [ISA-Tools](https://isa-tools.org/) also include a configuration for MIAPPE and can be used both for filling-in metadata and for validating.
   * [Collaborative Open Plant Omics (COPO)](https://copo-project.org/) is a data management platform specific for the plant sciences.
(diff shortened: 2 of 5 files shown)


bianchini88 (changes requested):
* "naming schema" looks a bit weird. Shall we have something like "naming convention"?
Or simplify and just say "consistent naming" ?
* What do we mean by "identifier schema" in the FAIRsharing entry? Also, why is this the only term which is not a plural?

pages/data_life_cycle/preserving.md
```diff
@@ -54,7 +54,7 @@ When preparing data for preservation several requirements need to be fulfilled.
   * Include information about provenance.
   * Include sufficient licensing information.
   * Ensure that data is well organised.
-  * Ensure that a consistent naming scheme is used.
+  * Ensure that a consistent naming schema is used.
```
nsoranzo, line 57: Not sure about the others, but "naming scheme" is much more used than "naming schema" (1.540.000 Google hits vs. 55.100).

floradanna: "naming convention" is what is meant here, I think

pages/your_domain/plant_sciences.md
```diff
@@ -108,7 +108,7 @@ It is recommended that metadata collection is contemplated from the start of the
   * [e!DAL](https://edal.ipk-gatersleben.de/) is a free data management platform for which MIAPPE templates are in development.
   * The [ISA-Tools](https://isa-tools.org/) also include a configuration for MIAPPE and can be used both for filling-in metadata and for validating.
   * [Collaborative Open Plant Omics (COPO)](https://copo-project.org/) is a data management platform specific for the plant sciences.
-  * [FAIRsharing](https://fairsharing.org) is a manually curated registry of reporting guidelines, vocabularies, identifier scheme, models, formats, repositories, knowledge bases, and data policies that includes many resources relevant for managing plant phenotyping data.
+  * [FAIRsharing](https://fairsharing.org) is a manually curated registry of reporting guidelines, vocabularies, identifier schema, models, formats, repositories, knowledge bases, and data policies that includes many resources relevant for managing plant phenotyping data.
```
floradanna, line 111: "scheme" might be the correct words here. Apparently, "identification scheme" seems to be a thing.

martin-nc:
Yes, looks like it's hard to come up with a simple rule. In an [earlier issue](https://github.com/elixir-europe/rdmkit/issues/1130) perhaps we were thinking of uses like "metadata schema" or "database schema".  But in cases like "naming scheme" then "scheme" is perhaps the better word, or "[naming convention](https://en.wikipedia.org/wiki/Naming_convention)". 

The [style guide](https://rdmkit.elixir-europe.org/style_guide) does not talk about all possible uses of scheme/schema, but just says, 'Use “schema” if you mean a pattern for structuring data.' I don't know if we need to go into more details here. It doesn't appear to be a common word on the site. 

So my vote would be to:

- Adopt "naming convention" instead of "naming schema/scheme".
- Keep "identification scheme" (if that's what it's [conventionally called](https://en.wikipedia.org/wiki/Identification_scheme)), and if the readers of the page will understand it.
- Stick with "metadata schema".

martin-nc:
@bianchini88 Yes - sorry, I wasn't paying enough attention to the actual sentence. It should definitely be plural. Should it be "identifier" or "identification"? On further (but not in-depth!) research, that seems more tricky to me now. They seem to be used interchangeably sometimes:

Screenshot 2023-02-09 at 17 10 17

"Identification scheme" seems to be used in cryptography and access control protocols (e.g. [this article](https://www.cambridge.org/core/books/abs/complexity-and-cryptography/identification-schemes/E64B768856B75A4AE06CCF042260E0D3)), but not really in the same sense.  I couldn't find anything on the FAIRsharing website. Maybe we should ask them? Other than that, and based on [this article](https://www.gov.uk/government/publications/linked-identifier-schemes-best-practice-guide/linked-identifier-schemes-best-practice-guide), I'd guess "identifier schemes" might be better. It's clearer that it's not about human identity, as well.

@nsoranzo yes, I was thinking the same. We should probably just remove "Schema" from the style guide, if everyone is happy with that. There isn't a straightforward way of covering all the uses of schema/scheme/convention etc.

bedroesb:
@nsoranzo @bianchini88 @floradanna and @martin-nc I think I covered the feedback in my latest commit, thanks a lot for looking into this!

bianchini88 (approved):

PR #1196, opened 2023-02-10
Removing "schema" from the style guide

martin-nc:
As [discussed](https://github.com/elixir-europe/rdmkit/pull/1194), it's too difficult to come up with a simple rule about where to use this word.


bedroesb:
Thanks martin!

bedroesb (approved):

PR #1206, opened 2023-02-24
Update brand names

bedroesb:


Nazeeefa (approved):
Thanks! We should add point about this in guidelines, with example that please make sure to write brand names properly such as GitHub not github (also make it helpful for screen readers to understand)

bedroesb:
Yes I agree! Or a general one, "if you use brand names, please check the correct spelling and use of capitals" or something alike.

PR #1218, opened 2023-03-23
More info on our contact page

bedroesb:


martin-nc (changes requested):
It should be "enquiry" if we're using British English. "Inquiry" is a formal investigation. American English just uses "inquiry" for both a formal investigation and informal questions.

Otherwise looks good!

bedroesb:
Oh thanks, again something I've learned! I also noticed I used capitals in the subheadings, took them away according to the style guide.

PR #1225, opened 2023-03-31

xiaoranzhou:
Added DataPLAN to the tool list. Not sure if this is the correct text location

### DIFF
--- a/pages/your_tasks/data_management_plan.md
+++ b/pages/your_tasks/data_management_plan.md
@@ -50,6 +50,7 @@ However, a number of web-based DMP tools are currently available that greatly fa
 * Choose one of the following online DMP tools (ordered alphabetically).
   * [Data Stewardship Wizard (DSW)](https://ds-wizard.org): publicly available open-source tool to collaboratively compose data management plans through smart and customisable questionnaires with FAIRness evaluation.
   * [DAMAP](https://damap.org/): tool for machine actionable Data Management Plans.
+  * [DataPLAN](https://plan.nfdi4plants.org): A FAIR and fast Data Management Plan generator for plant science.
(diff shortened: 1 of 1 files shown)


floradanna:
Thank you very much for your contribution. 

The editors will look into it.

xiaoranzhou:
> Thank you very much for your contribution.
> 
> The editors will look into it.

Thanks  a lot

Nazeeefa:
I guess this PR can be closed? as it will replace other PR that will involve adding DataPlan to Plant (your domain) page and Germany (national resources) page.

floradanna:
I would prefer to keep it open until the other changes are made (maybe by @wittigue and @sebeier), on the national page for Germany and plant sciences domain.

Nazeeefa (approved):

bedroesb:
@erlefloch What do you think of adding DataPlan to the plant page?

erlefloch:
@bedroesb currently the [plant sciences page](https://rdmkit.elixir-europe.org/plant_sciences) does not describe DMPs , I think it was meant to avoid duplicating information. 
ping @afablondon

floradanna:
It is true that a paragraph about DMP is missing in the plant page, but I think that it could be easily added since there is actually a DMP tool dedicated to plant research. A short paragraph will already be enough. However, if the tool is institute specific (it can only be used by specific institute/countries), it might be best to only describing it on the National page.
I don't know if it could be included in the plant tool assembly.

I guess that the other domains do not have a paragraph about DMP because they just use the generic DMP tools (DMPonline, DSW etc) mentioned in the generic task page, and/or institute specific template or questionnaire, mentioned in National pages.

sebeier:
Currently, it supports DMPs for Horizon 2020, Horizon Europe and the German BMBF and DFG. The main focus during development was to be able to be used with German funding agencies but was also extended to include other European funders.

erlefloch:
Draft PR : https://github.com/elixir-europe/rdmkit/pull/1246

erlefloch:
> Draft PR : #1246

now ready for merging I think

PR #1235, opened 2023-04-28
Revert "Update data_management_plan.md (#1225)"

bedroesb:
This reverts commit 4c4410926cb96e1aa629b4694a9b4a6452cac196.

Based on the conversation in #1225 I understood that dataPLAN will be part of the national resources of Germany, which is the case now. So this PR can be closed


bedroesb:
Based on the same conversation we might want to add it to the domain page plant though!

bedroesb:
This is getting further discussed in #1225

PR #1239, opened 2023-05-02
Rdmkit alliance

bedroesb:
This is the start of the RDMkit alliance page in the About


CaroleGoble (commented):
looks ok - but I kind of need to see the page....do we have a form for signing up?

bedroesb:
oh fair point! I will first start drafting the form!

bedroesb:
I will merge for sake of demonstartion!

PR #1240, opened 2023-05-05
add "metadata management" section to plant_phenomics_assembly.md

erlefloch:
It was in the WIP google doc since a while, but had never been pushed here.
Creation of "Metadata management" section with isa4j
Classification of dataverse, e!dal and zenodo under "Repositories" section


korbinib (approved):

pages/tool_assembly/plant_phenomics_assembly.md
korbinib, line 1: maybe it would be worth to update the graphic as well.
The tool is already in the table I see.

erlefloch: Good idea! Done

Issue #1114, opened 2022-12-08
Update request Plant sciences page


pinarpink:
Based on my quick review of the Plant sciences:

The tools PLAZA and PHIS are not described in the text but they exist in the tools table. They need to be added to the text. Also PIPPA is listed as a national resource for Belgium but not described in the Plants page. So  the generic PIPPA description could be added to the Plants page.

bedroesb:
Hey Pinar! This came to be because of the assemblies where these tools are part of, were added after the plant sciences page. But This problem will be solved on its own when we move to the new way of tagging tools

PR #1249, opened 2023-05-22
Moving to the new way of tagging tools in content

bedroesb:
Tools will be described in the main `tools and resources yaml file`:

image

And be linked from the content using 

`{% tool "tool-id" %}`

This will be visualized in the content like this:

image

Still on the TODO:

- [x] Replace tool links in the text with the new snippet
- [x] Add missing tools to the main yaml file
- [x] https://github.com/bedroesb/rdmkit/blob/demo/_data/tool_and_resource_list.yml replace
- [x] Adapt content so tools get mentioned at least once @ rdmkit
- [x] Agree on how to handle country specific tools
- [x] in national pages, change ref_to_main_resources so that it makes use of ID
- [x] update instance_of to make use of the tool ID
- [x] agree on (not) having backwards compatibility for old way of having related pages at tools/spreaadsheet
- [x] Update to spreadsheet convertor script to work with the yaml file so the yaml file gets updated with FAIRsharing/TeSS and Bio.tools IDs automatically
- [x] Nicer/better dropdowns for the tool information
- [x] Make the tools in tools table alphabetically.
- [x] Add new way of tagging the tools to the ELIXIR Toolkit Theme
- [x] Update documentation in how to add a new tool or resource
- [x] Make the snippet more forgiving
- [x] Adding a linter to check for:
  - [x] id is kebab-case
  - [x] Description does not contain colon
  - [x] URL works
  - [x] Make sure that id linked in "instance_of" exists in the main yaml file


bedroesb:
ERROR: Undefined tool ID: ELIXIR Core Data Resources in pages/your_tasks/existing_data.md

bedroesb:
XNAT assembly does not seem to have a working tools table

bedroesb:
This will also close #1114

martin-nc (approved):
Looks good to me!

bedroesb:
This is a big moment, lets gooo :rocket:

PR #1258, opened 2023-05-23
Update nels_assembly.md

smza:
changed the name SEEK to FAIRDOM-SEEK


smza:
Not needed. Name gets updated by the new tool snippet that is being used

PR #1337, opened 2023-07-10
Update main_tool_and_resource_list.csv

erlefloch:
removing data inrae from the upstream CSV to avoid it being brought back automatically like in https://github.com/elixir-europe/rdmkit/pull/1336


bedroesb:
thanks!

bedroesb (approved):

Issue #1276, opened 2023-05-25
Your Roles Pages


MijkeJetten:
**Link to the page of interest**
https://rdmkit.elixir-europe.org/your_role

**Describe what you would change to the page**
During the Ghent Contenthaton, we have been working on updates for the Your Roles pages, via this [Google Doc](https://docs.google.com/document/d/1NavSc77LsX4dSa7WrVF2TNCoNuL9byd3uviOJOZStGg/edit#). Some is still work in progress, but since we propose structural adjustments (headers, adding roles), it would be useful to see if the editorial agrees with the proposals in a broad sense. Happy to take along any suggestions, before we finalise, adjust with Markdown and submit.    

Contactpersons are me and Ulrike W.

**List the tool(s)or resource(s) that are missing on the page (if applicable)**
Not applicable

MijkeJetten:
To be discussed: add images for each page. We need help/suggestions for this to make sure it is not too much

MijkeJetten:
Ant thoughts by the editorial board before we publish a first page?

martin-nc:
FWIW I think they look good, and certainly better than what is there now. I'd say we go ahead and update the current pages/create the new ones. @wittigue what do you think?

MijkeJetten:
Ok with both new pages, looks good. One thing: shouldn't we update the contributors list as well to (alone) the Ghent Your Role authors?

Let's prep the other pages that are good to go as well --> PI & data steward. Can you help @wittigue @martin-nc?

MijkeJetten:
+ can we have a joint hour to finish the It & Trainer together, the three of us. ASAP if you are available @martin-nc @wittigue

PR #1338, opened 2023-07-14
New Policy maker role text

wittigue:
new Policy maker role text based on Google document from contentathon #1276

### DIFF
--- a/pages/your_role/data_steward_policy.md
+++ b/pages/your_role/data_steward_policy.md
@@ -1,7 +1,7 @@
 ---
 title: "Data Steward: policy"
 description: Data Steward with focus on data policies.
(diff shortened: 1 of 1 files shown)


floradanna (approved):
General feedback:
1. Revision of the list of links to training material: e.g. lessons about Data Carpentry might not be super relevant for DS policy officer. Maybe the list could be more focus on the policy.
2. About the paragraph "RDMkit pages", I was wondering if it was really necessary since almost all those page are listed as "related pages" from the metadata of the md page.

MijkeJetten:
For me the page is good to go. One thing: shouldn't we update the contributors list as well to (alone) the Ghent Your Role authors?

bedroesb:
This is a fair question @elixir-europe/rdmkit-editors any opinions?

MijkeJetten:
@bedroesb this is a fully new page, so any other contributors would be weird anyway. For the researcher your role page it might be somewhat different as there was a page before, but the updates are major (hardly any original text left) so I would go for the new contributors only for both pages

floradanna:
@MijkeJetten I am not sure it would be fair. I would keep the names of people that contributed to the previous pages anyway. The new pages are anyway written based on what was there before.

MijkeJetten:
Fine with everything, but we do need the add the new people if they are missing - do we then go for same contributor list for all your role pages, as it was a joint effort in the end.

bedroesb:
I will move it to the dev branch for these changes!

PR #1339, opened 2023-07-14
Apply typography suggestions

bedroesb:
### H1 updates
- [x] Default colour
### H2 updates
- [x] Keep Exo 2 
- [x] margin-top: 2.5 rem
- [x] margin-bottom: 0.5 rem
- [x] Font weight: ~500~ 600
- [x] Font-szie: 2.2 rem
### H3
- [x] font-weight : 600
- [x] Font-size: 1.5 rem

### DIFF
--- a/_sass/_bootstrap_variables.scss
+++ b/_sass/_bootstrap_variables.scss
@@ -27,6 +27,7 @@ $input-group-addon-bg: $light-x;
 $input-bg: $light;
 $input-border-width: 0;
 $h1-font-size: 1rem * 3.5 !default;
+$h2-font-size: 1rem * 2.2 !default;
 $h3-font-size: 1rem * 1.5 !default;
 $h4-font-size: 1rem * 1.1 !default;
 $h5-font-size: 1rem !default;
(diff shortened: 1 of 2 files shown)


bedroesb:
These are discussed changes from our monthly UX meetings

bedroesb:
definitely possible, the problem is a bit that we use font weights that are not always available with the fonts used by people...

bedroesb:
@martin-nc for example with the IDTk for me the jump is to big:
image

image

But this is just my opinion...

bedroesb:
Maybe I just have to get used to it :)

martin-nc:
Hmmm - I hadn't thought about people choosing different fonts for the toolkit theme. If that's the case, you might want to stick to 400 and 700 for that theme, and keep it simple. Each font should definitely have those! Unless it's a display or monospace font, of course, but you can't account for that. 

So then the questions is: do we stick to the same rules for the RDMkit? I don't think that's necessary. Once the theme is out there it's a free-for-all, and you can't enforce a law about how people should use fonts in it. For the RDMkit, I'd maybe make the `h2` 600, on reflection. It's not such a big change but it's more in keeping with the new `h3` rules, and it looks less clunky than the current 700. What do you think?

bedroesb:
@martin-nc I agree, I changed them to 600! I was thinking of merging these changes for now and doing rethinking the theme on its own another time :thinking:

PR #1347, opened 2023-08-07
No role tags for tools

bedroesb:
This goes together with #1346 #1344 #1338 and will take away the tools at the bottom of the pages


martin-nc (approved):
Thanks!

bedroesb:
Since currently this big list in the your role pages is of not much use, I will merge the page already!

PR #1349, opened 2023-08-09
New role page for Research Software Engineer

martin-nc:
Another page in the "Your role" section from the contentathon. See: https://github.com/elixir-europe/rdmkit/issues/1276.


bedroesb:
I will move this to the your role restructure branch, a new PRs will be opened where we can have a final look at the reorganization

PR #1360, opened 2023-08-16
Move to version 2.1.0 ELIXIR Toolkit Theme 

bedroesb:
This introduces:
- A table of content on the right side
- Cleaner lists in footer
- Truncate title in related pages all tools and resources


bedroesb:
This new version makes use of tool IDs which do not exist yet in the master branch of RDMkit...

PR #1369, opened 2023-08-28
Update de_resources.md

wittigue:
added related pages to ELN tools

### DIFF
--- a/pages/national_resources/de_resources.md
+++ b/pages/national_resources/de_resources.md
@@ -21,12 +21,21 @@ national_resources:
   - name: ELN Finder
     description: Electronic Laboratory Notebook (ELN) selection service. 
     url: https://eln-finder.ulb.tu-darmstadt.de/home
+    related_pages:
+      your_tasks: [metadata]
+      your_role: [researcher, research_software_engineer]
   - name: ELN Guide
     description: Guide for Electronic Laboratory Notebooks (ELN). 
     url: https://repository.publisso.de/resource/frl:6425772
+    related_pages:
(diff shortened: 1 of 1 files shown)


pages/national_resources/de_resources.md
```diff
@@ -21,12 +21,21 @@ national_resources:
   - name: ELN Finder
     description: Electronic Laboratory Notebook (ELN) selection service. 
     url: https://eln-finder.ulb.tu-darmstadt.de/home
+    related_pages:
+      your_tasks: [metadata]
+      your_role: [researcher, research_software_engineer]
```
bedroesb, line 26: There are not tools in the your_role pages anymore, so those will not show up :)

bedroesb: related pages only exists in the national tools, which currently only appear when also main tools are listed. So with the current mechanics, we first need to have main tools in the your_role pages if we want the national tools to show up there

wittigue: OK, so then we can close the PR because the ELN tools are not listed on any other page

bedroesb: Actually the metadata tag is in place is a good one imo, since there are main tools listed at the bottom, and the page talks about ELNs

bedroesb: Currently many other national tools still have their role tag, still waiting for their "faith", maybe we can it here also stay in place until we decide on tools on the your_role pages. I just wanted to make you aware that they currently would not show up :)

Nazeeefa (approved):

PR #1381, opened 2023-09-14
Update ethics.md

korbinib:


korbinib:
@Nazeeefa - added cross reference as suggested.

bedroesb (approved):

PR #1386, opened 2023-09-28
Update fi_resources.md

minnami:
I suggest IceBear to be removed from the list of tools. It is not mentioned anywhere in the text and after discussing with ELIXIR-FI I learned that it has some maintenance issues and thus it cannot be recommended at the moment.

### DIFF
--- a/pages/national_resources/fi_resources.md
+++ b/pages/national_resources/fi_resources.md
@@ -88,16 +88,7 @@ national_resources:
       your_role: [researcher, data_steward]
       your_tasks: [data_analysis]
     url: https://research.csc.fi/computing#cloud-computing 
-  - name: IceBear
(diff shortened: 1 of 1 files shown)


Nazeeefa:
Thanks @minnami I didn't fully understand - there are some other tools/resources that are also not mentioned in the text. On the website, it says new version is released https://icebear.fi/ Happy to approve and merge the PR, just want to make sure the page is updated correctly.

minnami:
Thank you for checking this @Nazeeefa Our node is currently reviewing our national resources & tool assembly pages so we'll be requesting some other changes too. In short: In this particular case we do not have a role in the development & maintenance anymore. Thus we would have to list & recommend dozens of other similar or similarly developed tools from universities in Finland if we include icebear.

Nazeeefa (approved):

PR #1394, opened 2023-10-04
Make CONTRIBUTORS more uniform

bedroesb:
@anamikachatterjee I was goin through the contributors and noticed you had an exceptionally long affiliation (compared to the others) Would you be ok in take out the department in the pull request? (see tab "Files changed")

### DIFF
--- a/_data/CONTRIBUTORS.yaml
+++ b/_data/CONTRIBUTORS.yaml
@@ -156,7 +156,7 @@ Emma Karoune:
 Graham Parton:
     git: gap736uk
     orcid: 0000-0003-4157-0352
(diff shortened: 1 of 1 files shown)


bedroesb:
Don't feel obliged of course, this is just a proposal!

bianchini88:
@bedroesb People do not usually add the department, so I would remove that instead of reducing the university name to its acronym.

bianchini88 (approved):

anamikachatterjee:
> @bedroesb People do not usually add the department, so I would remove that instead of reducing the university name to its acronym.

Sounds good to me :)

PR #1399, opened 2023-10-27
Update editorial_board_guide.md

korbinib:
added a note on suggestions


korbinib:
turns out we do not have octicons in EFF?

bedroesb (approved):

bedroesb:
@korbinib I added support for it, we can use these when we update our docuemntation about github. Thanks for the suggestion!

PR #1412, opened 2023-11-15
Move to new version of ETT

bedroesb:
This will bring following changes:

- "Tools and resources on this page" as heading
- Update Bootstrap to 5.3.2 and fix warning in build
- Only show pagination and search in datatables when relevant
- Add hide sidebar function
- Update countries automatically as it is now part of the theme
- Fix topnav highlight bug when toc: false


Nazeeefa:
> Add hide sidebar function

Do you mean the one on the left or the navigation pane in each sub-page?

bedroesb:
This is just a feature that the theme now decently supports. When no sidebar is specified, none will be given (the on one the left) and the content will take on the whole width of the page. In the past this would leave an empty gap.

bedroesb:
Not really relevant for the RDMkit since we always specify a sidebar

martin-nc (approved):
Look fine to me!

PR #1430, opened 2023-12-15
Update se_resources.md

YvonneKallberg:
Review of all links and resources, and update where necessary

### DIFF
--- a/pages/national_resources/se_resources.md
+++ b/pages/national_resources/se_resources.md
@@ -34,19 +34,19 @@ national_resources:
     related_pages:
       your_tasks: [dmp, data_publication, sensitive]
     url: https://nbis.se/support/supportform/index.php?form=consultation
-  - name: Swedish COVID-19 Data Portal
-    description: The Swedish COVID-19 Data Portal provides information, guidelines, tools and services to support researchers to utilise Swedish and European infrastructures for data sharing.
+  - name: Swedish Pathogens Portal
+    description: The Swedish Pathogens Portal provides information, guidelines, tools and services to support researchers to utilise Swedish and European infrastructures for data sharing.
     related_pages:
       tool_assembly: [covid-19]
       your_domain: [human_data]
       your_tasks: [sensitive, existing_data, data_publication]
-    url: https://covid19dataportal.se/ 
-  - name: SNIC 
-    description: The Swedish National Infrastructure for Computing (SNIC) is a national research infrastructure that makes available large-scale high-performance computing resources, storage capacity, and advanced user support, for Swedish research.
+    url: https://pathogens.se/ 
+  - name: NAISS 
+    description: The National Academic Infrastructure for Super­computing in Sweden (NAISS) is a national research infrastructure that makes available large-scale high-performance computing resources, storage capacity, and advanced user support, for Swedish research.
     how_to_access: An application is required to gain access to the compute and storage services.
     related_pages:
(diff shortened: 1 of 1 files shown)


YvonneKallberg:
I changed links to resources directly in the page, is this the way to do it or should it be somewhere more central in the main tool & resource page?

Nazeeefa:
@YvonneKallberg The links should also be mentioned/updated here in tools and resources list which can be found in `_data/tool_and_resource_list.yml`.

Nazeeefa:
Ja, only "The European Genome-phenome Archive (EGA)" is linked to the tools and resources table. I thought national covid-portals and national EGA instances are also considered tools/resources, that's what I was thinking of in my previous comment - e.g. I see FEGA Finland and Norway in Tools and Resources list https://rdmkit.elixir-europe.org/all_tools_and_resources and Sweden's Covid portal is also mentioned there https://covid19dataportal.se/

bedroesb:
@Nazeeefa this should not be the way of doing things. National instances are listed only on national pages, and not in the main tools and resources page. If this is the case, than this is wrong and this should be fixed.

bedroesb:
I migth have understand you wrong @Nazeeefa, I do not find FEGA back in https://github.com/elixir-europe/rdmkit/blob/master/_data/tool_and_resource_list.yml ?

bedroesb:
Some information: 
- the overview with all tools and resources in the RDMkit, https://rdmkit.elixir-europe.org/all_tools_and_resources, contains both the tools from the main yaml file and the ones listed in the country pages.
- If you update your national tools in the metadata of your national page, they will also be updated in the https://rdmkit.elixir-europe.org/all_tools_and_resources page
- In the National pages we describe resources tailored to users in that country (most of the time they also have a national domain)

bianchini88:
@bedroesb 
> I do not find FEGA back in https://github.com/elixir-europe/rdmkit/blob/master/_data/tool_and_resource_list.yml ?

The central EGA is indeed in the table (id: the-european-genome-phenome-archive)

The federated nodes should be on the national pages. 
On the NO page we have:

>  - name: Federated EGA Norway node

Which I believe is correct. A similar entry is present on the page for Finland, 
so I believe there is no issue with the current "positioning" of tools.

A question: the tools on national pages do not seem to have an "id" field. Is there a reason for that?

Nazeeefa:
> Some information:
> 
> * the overview with all tools and resources in the RDMkit, https://rdmkit.elixir-europe.org/all_tools_and_resources, contains both the tools from the main yaml file and the ones listed in the country pages.

Yes that's what I mean when I wrote "I see FEGA Finland and Norway in Tools and Resources list https://rdmkit.elixir-europe.org/all_tools_and_resources"
 
> * If you update your national tools in the metadata of your national page, they will also be updated in the https://rdmkit.elixir-europe.org/all_tools_and_resources page

Yep, makes sense!

YvonneKallberg:
Does this mean that I do not have to do anything additionally to what I have done? I've updated links in the metadata header of the national page, and that is it?

Nazeeefa (approved):

Nazeeefa:
@YvonneKallberg thank you! I think this is it, however I am wondering it might be useful to mention FEGA Sweden (https://fega.nbis.se/) in the metadata header of the page.

YvonneKallberg:
I've added a metadata post for FEGA-SE now

Nazeeefa (approved):

PR #1436, opened 2023-12-19
Update nels_assembly.md

bianchini88:
Using a country keyword for a tool assembly might create undesired results when courses for other independent national resources/tools are registered into TeSS. 

With the current suggestion some results of the query get lost, but maybe the annotations on TeSS should be improved (e.g. including explicitly NeLS as a keyword or similar).

### DIFF
--- a/pages/tool_assembly/nels_assembly.md
+++ b/pages/tool_assembly/nels_assembly.md
@@ -10,7 +10,7 @@ related_pages:
 training:
   - name: Training in TeSS
     registry: TeSS
-    url: https://tess.elixir-europe.org/search?q=norway
+    url: https://tess.elixir-europe.org/search?q=NeLS+Norway
 ---
 
 ## What is the NeLS data management tool assembly?


bedroesb (approved):

Nazeeefa:
@bianchini88 I think https://tess.elixir-europe.org/search?q=norway should stay as it gives all results with all three types (events, materials, and content providers), and more keywords should be added to the https://tess.elixir-europe.org/materials/using-the-norwegian-e-infrastructure-for-life-science-and-usegalaxy-no instead. We should also use https://github.com/ELIXIR-Norway-Training to share materials from trainings other than DMPs, and mention it in this page under training.

bianchini88:
@Nazeeefa No, it should not stay, as it will provide results e.g. the Omnipy Workshops that are not relevant to NeLS. In general, it would be wrong to use a country word to look for training related to a specific tool, it just would not scale right...

korbinib (approved):

Nazeeefa:
Sorry I got confused for a second and thought you are working on Norway's national resources page.

> Why not on the national page? In general, I prefer using TeSS materials when this links to a Zenodo entry. We should possibly have releases from the GitHub repo on Zenodo and link them from TeSS (we are possibly doing this already, aren't we?)

Yes I meant national page. I prefer that too but I think it's important to make GitHub repo visible to our audience too, and also mention it in our TeSS entries. Material we upload on GitHub (at least in terms of our DMP workshops) is also uploaded to Zenodo and get linked in TeSS. We can take this discussion on Slack, given that this is merged now :)

PR #1453, opened 2024-01-25
Update news.yml

bianchini88:
Added linked PR to new item


bedroesb:
@bianchini88 I don't think it is relevant to link the PR in this news item since it is not linked to work that was done for the newsitem. (there is no added value/content in the pull request)

bianchini88:
@bedroesb I do not disagree with what you are saying. It is just that from the graphical point of view it looks like someone forgot to add that piece of information (maybe it is only in my head).

Nazeeefa (approved):

Nazeeefa:
I approved because it is now consistent with how other news items about new pages are published.

bedroesb:
I don't have a strong enough opinion to block this one ;)

PR #1454, opened 2024-01-25
Update references.bib

korbinib:
fixes the author list. I would strongly suggest to document a workflow for future applications using tooling (e.g. zotero) to avoid the issue.

### DIFF
--- a/_bibliography/references.bib
+++ b/_bibliography/references.bib
@@ -1,6 +1,6 @@
 @article{field2008migs,
   title={The minimum information about a genome sequence (MIGS) specification},
-  author={Field, D. et al.},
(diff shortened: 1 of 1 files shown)


bianchini88:
@korbinib All good, but we just merged the ML domain page which comes with its own set of bibtex entries. If you share the relevant links/documentation I can try to fix those entries in the same way you fixed these.

korbinib:
Bad timing I guess.

Steps (4.-5. optional):

1. Get https://www.zotero.org/
2. Get better BibTeX https://retorque.re/zotero-better-bibtex/installation/index.html
3. Use auth.lower + year + shorttitle(1,0)  for citation key generation (old entries should be pinned)
4. Get an account
5. Add this group https://www.zotero.org/groups/5371154/rdmkit
6. Add/Import items to zotero (optional to group above)
7. Export as better BibTex and append to references.bib

bianchini88:
Thank you! I will give it a go.

bianchini88:
@korbinib @bedroesb is there any particular reason for this being still at the *draft* stage? 
I checked on a fork and did not encounter any conflict.

korbinib:
Please review and merge @bianchini88

bianchini88 (approved):

PR #1448, opened 2024-01-19
Improved bibliography

bedroesb:
This should fix the remark of @bianchini88  in #1447

PR #1456, opened 2024-01-26
changing bibliography style to nature

bianchini88:
Changing the bibliography style to *Nature* as suggested by @martin-nc in #1448.

### DIFF
--- a/_bibliography/references.bib
+++ b/_bibliography/references.bib
@@ -132,7 +132,7 @@
 
 @article{field2008migs,
   title={The minimum information about a genome sequence (MIGS) specification},
-  author={Field, D. and et al.},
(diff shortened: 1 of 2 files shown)


bianchini88:
While this works and helps to standardise the bibliography (only initials and not a mix and match of initials and full names when available on the bib file), it introduces an additional list of numbers, see the screenshot below:
image

This is apparently a very common issue, see https://github.com/inukshuk/jekyll-scholar/issues/203
although it does not seem there is one clear and simple solution.

This should probably be investigated and discussed by people more competent than me in Jekyll.

bedroesb:
```css
ol.bibliography {
    list-style-type: none;
    padding-left: 0;
}
```

makes the pink one disappear:

image

But is off course less pretty

bedroesb (approved):
Looks good now!

PR #1470, opened 2024-03-25
Update Dutch national resources page

rnavest:
Not 100% sure whether the links added for zenodo and fairsharing are in the correct format as there was no example for this in the description https://rdmkit.elixir-europe.org/page_metadata


bedroesb:
You are right, we should document better the supported registries when listing national tools... The registries that can be used are listed in our https://rdmkit.elixir-europe.org/tool_resource_update page. Zenodo is not one of them so far...

rnavest:
@bedroesb as cBioPortal is an international tool and not strictly tailored to users in the Netherlands, do you think it would be better to move it to the "Netherlands contributes to" section of the national page (https://rdmkit.elixir-europe.org/nl_resources)? In that case I could add cBioPortal to the tool_and_resource_list.yml file, like e.g. FAIRsharing or XNAT, as described in the link you shared.
For the training that I would like to reference, I'll check whether I'm allowed to upload it to TeSS so it can be referenced.

bedroesb:
@rnavest when I look at the resource, I think you are right, it does belong to the general tools and resources table. The only requirement for this to happen, is that the tool and resource must be described/mentioned somewhere in our content (for example task or domain page). Would you be willing to do this?

PR #1475, opened 2024-04-15
Improve accordion styles

martin-nc:
Currently, the last accordion drop-down in the related links section does not have a bottom border. See, for example, "Tools and resources on this page" here: https://rdmkit.elixir-europe.org/data_brokering. This PR adds a border to the last item. It also removes the top border hover style (the border darkens when you hover over the accordion item). Instead, it changes the colour of the accordion heading to blue when you hover over it.

**Before:**
Screenshot 2024-04-15 at 15 06 00

**After:**
Screenshot 2024-04-15 at 15 25 30


bedroesb:
Thanks a lot! How does it look like when expanded? Is the border under the content of the collapsible?

martin-nc:
@bedroesb Yes, the border applies to the header container only when it is collapsed:
`a.info-collapse[aria-expanded="false"]:last-of-type`
and applies to the info text when expanded:
`a.info-collapse[aria-expanded="true"]:last-of-type + .show`

https://github.com/elixir-europe/rdmkit/assets/15261745/2fce458c-0b90-4210-bf5d-eb4a7d0a6df8

bedroesb (approved):
@martin-nc This is great! I just went through the pages and everything looks to behave as it should! I should add this to the Toolkit Theme ;)

bedroesb (approved):

PR #1481, opened 2024-04-30
Update working_with_git.md

### LINKED ISSUE #1480: command for git cloning the repository fails
**Link to the page of interest**
https://rdmkit.elixir-europe.org/working_with_git#forking---branching---changing---pushing---pr

``` git clone git@github.com:USERNAME/rdmkit.git``` does not seem to be working for me.
I would recommend changing it with ``` git clone https://github.com/USERNAME/rdmkit/ ```.
If the former works for somebody, we could also display both options.

bianchini88:
Closes #1480


bedroesb:
@bianchini88 Maybe add to this that the current way is using the recommended SSH way of cloning a repo, and that for https you need to follow different commands.

PR #1484, opened 2024-05-06
Update DSW API URL

MarekSuchanek:
We recently migrated to a new domain name under elixir-europe.org ... Currently, there is a redirect which should work well; however, that will be removed eventually.


diana-pilvar (approved):

bianchini88:
I started clicking through the new links and realised that the LS login does not seem to be working on the CONVERGE instance (i.e. https://converge.dsw.elixir-europe.org). Also, after a quick look, I see that some of the content is hidden behind login, which is not ideal for the average user/researcher. We might have to rethink some of the links slightly.

MarekSuchanek:
> I started clicking through the new links and realised that the LS login does not seem to be working on the CONVERGE instance (i.e. https://converge.dsw.elixir-europe.org). Also, after a quick look, I see that some of the content is hidden behind login, which is not ideal for the average user/researcher. We might have to rethink some of the links slightly.
> 
> (this said, I have nothing against merging the PR)

Thanks @bianchini88, LS Login fixed now... We can discuss what should be more exposed there elsewhere. Can be done through making it publicly available and add links e.g. to a customized login page 🤷🏻‍♂️

bedroesb:
@MarekSuchanek for me the LS login is not working at https://researchers.dsw.elixir-europe.org/wizard/. I was wondering if the observations you made @bianchini88 are also applicable for https://researchers.dsw.elixir-europe.org/wizard/ ? since that is the one that will be in place instead of the converge one. Correct me if I am wrong!

MarekSuchanek:
> @MarekSuchanek for me the LS login is not working at https://researchers.dsw.elixir-europe.org/wizard/. I was wondering if the observations you made @bianchini88 are also applicable for https://researchers.dsw.elixir-europe.org/wizard/ ? since that is the one that will be in place instead of the converge one. Correct me if I am wrong!

You are never wrong! 😄 I fixed it also there... thanks for noticing.

bianchini88:
> LS Login fixed now... 
Thank you, @MarekSuchanek 

> I was wondering if the observations you made @bianchini88 are also applicable for https://researchers.dsw.elixir-europe.org/wizard/ ? since that is the one that will be in place instead of the converge one. Correct me if I am wrong!

Hi @bedroesb, I think the deep links are working as intended. I was more concerned about e.g. the tool 
```
 id: dpia-knowledge-model
 name: DPIA Knowledge Model
 url: https://converge.dsw.elixir-europe.org/knowledge-models/elixir.lu:dpia-research:0.1.0
```
used [here](https://rdmkit.elixir-europe.org/gdpr_compliance#how-do-you-protect-research-data-under-gdpr) at the bottom (DPIA Knowledge Model). Note also that this particular Knowledge Model is not available at https://researchers.dsw.elixir-europe.org/wizard/, and we probably would not want people to jump from different instances...
but I agree with @MarekSuchanek on the fact that this should probably be discussed elsewhere.

floradanna (approved):

PR #1489, opened 2024-05-22
Start implementing new icons

bedroesb:
We have been treated with wonderful new icons by Xenia. (Some are still on the way) But we can have a first look and feel here:

https://bedroesb.github.io/rdmkit/


bedroesb:
@martin-nc thanks! Could you approve the PR? Than it can go live together with the new icons on our poster!

bedroesb:
This shows that working from a mobile does not work well 😅 sorry!

PR #1494, opened 2024-05-31
Clarify steps in local Jekyll run

GilOliveira:
More specifically:
- Add the Bundler install command
- Separate dependency install step
- Small writing tweaks

The motivation is I was installing Jekyll locally on my machine and it was erroring out because the Bundler install command was missing from the instructions 😅


bedroesb (approved):

bedroesb:
it used to be there but for some reason it got taken out, thinking it was not needed (My fault I think!)

bedroesb:
Thanks!

PR #1495, opened 2024-05-31
Add elabFTW to tools list

GilOliveira:
[elabFTW](https://www.elabftw.net/) is a FOSS ELN.

### DIFF
--- a/_data/tool_and_resource_list.yml
+++ b/_data/tool_and_resource_list.yml
@@ -2473,6 +2473,8 @@
 - description: sceasy is a package that helps easy conversion of different single-cell data formats to each other.
   id: sceasy
   name: SCEasy
+  registry:
+    biotools: sceasy
   url: https://github.com/cellgeni/sceasy
 - description: Interactive portal for single-cell genomics data.
   id: single-cell-portal


GilOliveira:
Converted to draft as it looks like my editor is changing the single quotes to double quotes in other entries...

GilOliveira:
Okay, ready to go!

GilOliveira:
> all tools need to be put in context on the RDMkit by describing it within the text.
@bedroesb My objective in adding the tool is I'm working on a PR to update the Portuguese resources page, and we have an instance of elabFTW running, and wanted to link back to a tool page using the `instance_of` property. Would that suffice?

bedroesb:
Sadly you would still have to add it somewhere in a relevant guideline page because that is the rule in other to get it in the main tools table. But indeed once that's done, you can this way describe an instance of the tool in the national page.

bianchini88:
@GilOliveira consider adding a TeSS query to this entry, which would bring people to your own workshop material:
https://tess.elixir-europe.org/materials/ready-for-biodata-management-intensive-course

bedroesb:
@GilOliveira Any news on referencing the tool inside of the RDMkit?

PR #1496, opened 2024-05-31
Add community and RDMkit CLub page

bedroesb:
image

image

### DIFF
--- a/pages/about/community.md
+++ b/pages/about/community.md
@@ -10,8 +10,8 @@ The RDMkit community aims to foster the adoption, enhancement, and incorporation
 The community convenes under the RDMkit Club, a platform for collective engagement where members share experiences, address queries, troubleshoot issues, and stay updated on pertinent developments. Additionally, we maintain close collaboration with the [ELIXIR RDM Community](https://elixir-europe.org/communities/research-data-management).
 
 RDMkit community includes:
-- **[Contributors](contributors)**
-- **[Editorial Board](editorial_board)**
-- **[RDMkit Club](rdmkit_club)**
+- [**Contributors**](contributors)
+- [**Editorial Board**](editorial_board)
+- [**RDMkit Club**](rdmkit_club)
 
 
--- a/pages/about/community.md
+++ b/pages/about/community.md
@@ -6,7 +6,7 @@ title: Community
 Welcome to the RDMkit community! 
 
 
-The RDMkit community aims to foster the adoption, enhancement, and incorporation of Research Data Management (RDM) best practices collected on the RDMkit. Initially established as a component of the ELIXIR Converge project, the RDMkit community has experienced continuous growth, an esteemed editorial board, committed contributors, and a growing user base.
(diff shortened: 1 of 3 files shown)


pages/about/community.md
```diff
@@ -0,0 +1,17 @@
+---
+title: Community
+---
+
+	
+Welcome to the RDMkit community! 
+
+
+The RDMkit community aims to foster the adoption, enhancement, and incorporation of Research Data Management (RDM) best practices collected on the RDMkit. Initially established as a component of the ELIXIR Converge project, the RDMkit community has experienced continuous growth, an esteemed editorial board, committed contributors, and a growing user base.
```
bianchini88, line 9:
```suggestion
The RDMkit community aims to foster the adoption, enhancement, and incorporation of Research Data Management (RDM) best practices collected on the RDMkit. Initially established as a component of the ELIXIR-CONVERGE project, the RDMkit community has experienced continuous growth, an esteemed editorial board, committed contributors, and a growing user base.
```

pages/about/rdmkit_club.md
```diff
@@ -0,0 +1,20 @@
+---
+title: RDMkit Club
+---
+
+RDMkit Club is an informal drop-in session held on the **last Thursday of each month at 10:00 CEST/ 9:00 BST**, where we come together and discuss all topics related to RDMkit including its impact and future directions.
```
bianchini88, line 5: > at 10:00 CEST/ 9:00 BST**

Do we change these when the summertime is over? Use UTC?

bedroesb: Fair point, if the online calculators are correct this will be 8h UTC, but does this solve the problem? Because summer or winter, it will be at 10h / 9h ?

bianchini88: It is my understanding that UTC does not change depending on the time of the year, so that would solve the problem, I believe.

bedroesb: The fact that it does not change, does that not create the problem? Sorry I feel stupid about this :D Because if CET is UTC +1 and CEST UTC +2, than it means with a fixed UTC hour, the hour of the meeting changes winter vs summer?

smza: The UTC can be very confusing!

bianchini88: > The UTC can be very confusing!

Agreed, shall we use "CE(S)T" instead

bianchini88: @bedroesb Thinking about this, I suspect you are right.

martin-nc: I think the 10:00 CEST/ 9:00 BST option is more intuitive as well. Even if we forgot to change it, I think I'd find it easier to convert from BST to GMT than from UTC to whatever time zone we're in! It's what you're used to, I guess.

bedroesb: @martin-nc What do you think of the CE(S)T solution, are is it abvious enough?

martin-nc: I think so. Would you miss out BST/GMT then?

bedroesb:
Fair enough, now I only need an approval :smile:

martin-nc (approved):
Okay, let's see how that time format goes!

bianchini88 (approved):

bedroesb:
Thanks for the reviews!

PR #1501, opened 2024-06-10
Fix visual artifact fix for high dpi screens

bedroesb:
before: 

image

After:

image

### DIFF
--- a/_sass/_custom_classes.scss
+++ b/_sass/_custom_classes.scss
@@ -48,6 +48,8 @@
 }
 
 #main {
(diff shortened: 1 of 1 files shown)


martin-nc (approved):
I can't seem to replicate that, but maybe because my screen isn't good enough! The change looks fair enough, though.

bedroesb:
@martin-nc go to a page with not a lot of content, like your role, and zoom out to mimic a 4k screen that has the scaling "zoomed out"

PR #1506, opened 2024-06-20
Add new link to template slides

bedroesb:
These are the template slides created by Xenia

### DIFF
--- a/pages/about/media_kit.md
+++ b/pages/about/media_kit.md
@@ -68,9 +68,11 @@ For use in presentations, tutorials and all other RDMkit related activities the 
 
 ## Presentations
 
-#### RDMkit in 1 and 3 slides
+#### RDMkit template slide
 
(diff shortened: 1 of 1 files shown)


martin-nc:
The first two slides look a bit odd to me, like the screenshot has accidentally got moved off the page. Is that intentional? Maybe I haven't understood how it will be used.

bianchini88:
I share Martin's concern about the first two slides. Also, in the 3rd one I read "16 current editorial board members" but I see only 15 pictures/names...

lauportell:
I thought the same at the beginning but try to show the slides in presenter mode and it looks very cool!

marikapop:
I would delete a cookie banner on the screenshot on second slide.

bedroesb:
@bianchini88 and @martin-nc like Laura mentioned, it is to be seen in presenter mode, and it is a feature not a bug ;)

GilOliveira:
Love the style and the icons!!
Shouldn't the transition between slides 1-2 be a slide down to revel the rest of the screenshot, instead of a slide right as it is right now?

martin-nc:
I thought a bit the same as @GilOliveira regarding the transition, but I don't think there is a "slide down" option. Still, I'm happy to go with the popular vote on this!

GilOliveira:
I thought a bit the same as @GilOliveira regarding the transition, but I don't think there is a "slide down" option.Oh! I figured it was something that was bound to exist… in that case, I’m all for it! Looks great!

bedroesb:
What if we just take the animation away? ;) Sorry I was not aware the weirdness came from the transition.

bedroesb:
I normally have implemented the feedback and created alternative Opening slides. If somebody can review this, than this can go life!

martin-nc (approved):
Yes, people have a choice of starts now.

PR #1532, opened 2024-09-26
Improve the concept of "resources" in the RDMkit

### LINKED ISSUE #1518: A clearer definition of "resources"
**Description**

Refer to https://rdmkit.elixir-europe.org/tool_resource_update

Whilst it is understandable what a tool is, the term "resource" is considerably more vague and open to interpretations.
The possibility of linking with FAIRsharing would make one think that "standards", "databases", and "policies" are all to be considered as "resources" and should be in the yml file. In the case of standards, this is implemented e.g. for MINSEQE 
[here](https://rdmkit.elixir-europe.org/metadata_management#how-do-you-find-appropriate-standard-metadata-for-datasets-or-samples). The same approach is, however, not implemented in many other pages (e.g. MIxS [here](https://rdmkit.elixir-europe.org/human_pathogen_genomics#planning-a-study-with-pathogen-genomic-data)).  

Another interpretation of the FAIRsharing linking would be to use it for "standards", "databases" and "policies" which are supported by the various "tools" or "resources", in which case metadata standards should not be in the yml file. I am sure other interpretations are also possible.

Personally, I am in favour of the first interpretation, as it would provide richer information and the possibility of linking towards both the webpage of the "resource" and the registries (FAIRsharing, bio.tools, TeSS). One could also consider having other categories that behave as "tools" but use other icons (it would be weird to have that wrench icon close to e.g. a database or policy).

Independently from the paragraph above, we should clarify and be more explicit about what is a "resource" (i.e. what has to go in the yml file and what does not).

bedroesb:
This will close #1518 
- [x] Choosing a more neutral icon for the tools in the text
   image
- [x] Define what resources are with examples in our contribute documentation


bianchini88:
@bedroesb I do not think it is up to me to approve the new icon. If we really want to close https://github.com/elixir-europe/rdmkit/issues/1518 we should also expand the documentation. I can have a go at it, but not before the 17th of October.

bedroesb:
I agree, this needs more thought than just the icon.

bedroesb:
Based on the editorial meetings, an info icon fits best the purpose of the clickable 'buttons'. 
Two new variants:
image
image

bianchini88:
I do like the ones with the colour filling and intermediate "i" sizes.

bedroesb:
@bianchini88 I've added some numbers, makes it easier :) I am also adding a paragraph to the 'Update a tool or resource''

bianchini88:
Thanks @bedroesb, I actually struggled to understand how many variations we had. I would say 3>1>2>4 (but please ask someone with a better graphic sense than me).

bedroesb:
I added 2 more, my current favorites are 3 and 6

bedroesb:
@elixir-europe/rdmkit-editors I added 6 options for an icon above and updated the **Add a new tool or resource** contribute page be more clear about when a tool should be added to the tools table. Did I forget things or remember things wrongly? Please do give feedback. I might also use the word `main YAML file` to much ....

Nazeeefa:
I am afraid I can't vote for any of the icon options for tools and resources because I like the current gear icon 😄 The "i" icon suggests, to me at least, that hovering over icon will give me a brief information and I do not expect anything more.

diana-pilvar:
if only used for tools, I like the gear. If the links also cover resources, gear is not optimal. 
I don't like the box.
For the I's : 4,3 >1>2,4
6 reminds me of biohazard

pages/contribute/tool_resource_update.md
```diff
@@ -35,15 +50,11 @@ Example:
...
 
+## Linking the tool or resource in the text with the main YAML file
 
-## What tool or resource can be added to the table
-Tools and resources specifically mentioned in the text of the pages should be present in the main table. 
-
-## Making changes
-
-1. Make sure the tool you want to add is not already described in the [yaml file](https://github.com/elixir-europe/rdmkit/blob/master/_data/tool_and_resource_list.yml). If so, go to step 3, if not, follow the next step.
+1. Make sure the tool you want to add is not already described in the [YAML file](https://github.com/elixir-europe/rdmkit/blob/master/_data/tool_and_resource_list.yml). If so, go to step 3, if not, follow the next step.
```
bianchini88, line 55:
```suggestion
1. Make sure the tool you want to add is not already listed in the [YAML file](https://github.com/elixir-europe/rdmkit/blob/master/_data/tool_and_resource_list.yml). If so, go to step 3, if not, follow the next step.
```

bianchini88 (changes requested):

korbinib:
1>3>2>4

bedroesb:
In the editorial meeting we voted on icon number 3. This is put in place now. @bianchini88  I think I incorporated all your feedback, what do you think?

bedroesb:
@elixir-europe/rdmkit-editors The new logo is in place and our documentation about what is a resource is improved. This PR is ready I think!

bianchini88 (approved):

PR #1543, opened 2024-10-11
Create a  CITATION.cff

### LINKED ISSUE #1529: CITATION.CFF file
**Description**
RDMkit already has a citation requirement: https://rdmkit.elixir-europe.org/about#how-to-cite-the-rdmkit

To allow proper and easy attribution of the resource, [CITATION.cff file](https://citation-file-format.github.io/) can be added.

bedroesb:
this will close #1529


CITATION.cff
diana-pilvar, line 1: maybe space after : ? RDMkit: The Research.....

CITATION.cff
```diff
@@ -0,0 +1,13 @@
...
+- family-names: "ELIXIR"
+title: "RDMkit:The Research Data Management toolkit for Life Sciences"
+version: 2.0.0
+doi: 10.5281/zenodo.5110060
+date-released: 2021-03-01
+url: "https://github.com/elixir-europe/rdmkit"
+preferred-citation:
+  type: generic
+  authors:
+  - family-names: "ELIXIR"
```
vildead, line 13:
```suggestion
    - family-names: "ELIXIR"
  year: 2021
  title: "Research Data Management Kit. A deliverable from the EU-funded ELIXIR-CONVERGE project (grant agreement 871075). URL: https://rdmkit.elixir-europe.org"
```

This will render the string almost the same as the one on our About page:
 `ELIXIR. (2021). Research Data Management Kit. A deliverable from the EU-funded ELIXIR-CONVERGE project (grant agreement 871075). URL: https://rdmkit.elixir-europe.org/`

CITATION.cff
```diff
@@ -0,0 +1,13 @@
+cff-version: 1.2.0
+message: "If you use this software, please cite it as below."
+authors:
+- family-names: "ELIXIR"
```
vildead, line 4:
```suggestion
  - family-names: "ELIXIR"
```

CITATION.cff
```diff
@@ -0,0 +1,13 @@
...
+title: "RDMkit:The Research Data Management toolkit for Life Sciences"
+version: 2.0.0
+doi: 10.5281/zenodo.5110060
+date-released: 2021-03-01
+url: "https://github.com/elixir-europe/rdmkit"
```
vildead, line 9:
```suggestion
repository-code: "https://github.com/elixir-europe/rdmkit"
url: "https://rdmkit.elixir-europe.org"
```

CITATION.cff
vildead: This is really optional as it adds weird dots.

Otherwise, we dont need the whole `preffered-citation` block I guess

vildead (changes requested):
few suggestions

CITATION.cff
bedroesb: Actually, we should update the one in the about!

CITATION.cff
bedroesb: done!

CITATION.cff
```diff
@@ -0,0 +1,16 @@
+cff-version: 1.2.0
+message: "If you use this software, please cite it as below."
```
bianchini88, line 2: Why "software"?

bedroesb: @bianchini88  Good catch, What about "If you use content of this website, please cite as below."?

bianchini88: @bedroesb Maybe something like "When you use the information provided on this website, please cite as below". Just to make sure it is about the information on the pages and not only the text itself.

bedroesb: @bianchini88 not sure which ones is better if I ma honest. Other opinions? @elixir-europe/rdmkit-editors 

```
If you use content of this website, please cite as below
```
vs
```
When you use the information provided on this website, please cite as below
```

Nazeeefa: I prefer latter "When you..", it's more welcoming :)

bedroesb: I agree! And if it is:

```
When you use content of this website, please cite as below
```

Nazeeefa: Sounds good to me! I see the difference between the use of "content of.." vs "information provided.." and I prefer former ("content of..") :)

bianchini88: Just to clarify, I suggested moving away from "using content" because that could be interpreted as "if you are using the actual text" (i.e. if you quote) and that is probably not what we want. On the other hand, I do appreciate that the "content of" sentence flows a bit better. Another possible solution would be to move away from the word "use" entirely. Something like ``` When referencing this website, please cite as below ```. Anyway, I am probably overthinking this.

Nazeeefa: I don't think you are overthinking, I appreciate your view @bianchini88  How about "When using content from the RDMkit pages, please reference the website by citing it as follows:" ?

bianchini88: @Nazeeefa, it is a better phrasing, but the ambiguity in the word "using" is still there. We inherited this word from the initial commit, which contained a sentence that was software-related, and one does not "use" RDMkit in the same way one would "use" a piece of software. I think the most correct verb to substitute for "using" would be "referencing" in this case (but feel free to suggest some others). I would suggest ```When referencing content from the RDMkit pages, please cite as follows:```

bedroesb: I like that version!

bedroesb:
image

PR #1552, opened 2024-10-29

bianchini88:
Fixing wrong FAIRsharing ID for a resource (which currently brings up a 404).

The tool is mentioned [here](https://rdmkit.elixir-europe.org/omero_assembly#what-is-omero)  before the figure and leads to https://fairsharing.org/FAIRsharing.4747 (404).


bedroesb (approved):
well spotted!

bedroesb:
I checked an now it is indeed correct.

PR #1555, opened 2024-10-29

bianchini88:
Changing URL to the GitHub page instead of having a duplicate link to bio.tools. The GitHub page URL is taken from the bio.tool entry https://bio.tools/plant-3d

### DIFF
--- a/_data/tool_and_resource_list.yml
+++ b/_data/tool_and_resource_list.yml
@@ -58,7 +58,7 @@
   name: Plant 3D
   registry:
     biotools: plant-3d
-  url: https://bio.tools/plant-3d
+  url: https://github.com/iziamtso/P3D/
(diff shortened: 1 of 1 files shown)


bedroesb (commented):
I am personally also a bigger fan of linking to GitHub pages

bedroesb (approved):

bianchini88:
> I am personally also a bigger fan of linking to GitHub repos

This would open the discussion of what to pick if you have a website and a GitHub repository. Maybe we need a specific field for GitHub repos? Something to discuss at the RDM community event?

bedroesb:
GitHub repos tend to be more persistent. Don't know if having it separate is needed. Keeping track of 2 urls sounds like something bio.tools/fairsharing should take care of. Happy to discuss it in Luxembourg!

PR #1574, opened 2024-12-06
Change title to Creating a data workflow diagram

bedroesb:
THe current title is not reflecting the content of the page

### DIFF
--- a/_data/news.yml
+++ b/_data/news.yml
@@ -194,7 +194,7 @@
   date: 2024-10-24
   linked_pr: 1526
   description: The page is enriched with additional links and a bibliography. [Discover the page here](human_pathogen_genomics).
-- name: "New page: Creating a data workflow"
+- name: "New page: Creating a data workflow diagram"
   date: 2024-12-06
   linked_pr: 1528
   description: This page guides you with capturing your planned data workflow in a diagram. [Discover the page here](creating_data_workflow).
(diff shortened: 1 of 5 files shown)


bedroesb:
@vildead

Nazeeefa:
@bedroesb Thanks for suggesting the title ‘Creating a workflow diagram’! :) To refine it, it would be helpful if @vildead can kindly clarify if a diagram is suggested to represent a flow of project data or structured flow of data analysis? Based on the page content, it seems former (not data analysis workflows in a bioinformatics context), so incorporating ‘project’ in the title might be more precise. Easier and better approach to this is mentioning clearly in page text the definition of workflow in a context of data management (to avoid confusion with bioinformatics definition). Sorry for any inconvenience from my side.

marikapop (approved):

bedroesb:
@Nazeeefa I agree that this is indeed some context that would be nice to have! I will go ahead and merge this to make sure the current page url does not get used to much :)

bedroesb:
@marikapop thanks!

bedroesb:
Or wait I read to quickly, @ nazeefa, would you chaneg the title to Creating a data project diagram?

marikapop:
> Or wait I read to quickly, @nazeefa, would you change the title to Creating a data project diagram?

Now I do not like this title :) It sounds to me like Creating a diagram of a data project.
What about Creating a project dataflow diagram or simpler Creating a dataflow diagram.

bedroesb:
@marikapop I like that last one!

bedroesb:
@marikapop Could you check with @vildead if he is ok with this change?

vildead:
nice! Thank you for the changes. Indeed, its more about making a diagram

PR #1575, opened 2024-12-09
lychee version format

korbinib:


bedroesb (approved):

bedroesb:
sloppy on my side, this works indeed!

PR #1587, opened 2025-01-17
Replacing outdate form of FAIRsharing IDs

bianchini88:


bianchini88:
@bedroesb I started the process here.

bianchini88:
@bedroesb I had a quick look locally and it seems to be working for collections as well. I will open this for review.

bedroesb (approved):
Looks good! DO we need to do something else?

PR #1590, opened 2025-01-17
Update tool_resource_update.md

### LINKED ISSUE #1553: FAIRsharing IDs for collections
**Link to the page of interest**
https://rdmkit.elixir-europe.org/tool_resource_update#the-main-yaml-file

> FAIRsharing collections have an ID that follows the pattern bsg-s000XXX.

This does not seem to be the general case, see for example https://fairsharing.org/3505 or https://fairsharing.org/3508, 
these collections use only the four digits as ID, and I cannot locate anything starting with "bsg". 

Note also that the ID for COPO in the table (https://fairsharing.org/bsg-d001247) does not bring to a collection but to an entry of the Database registry, which actually has ID FAIRsharing.91a79b (although bsg-d001247 also works, but is not even displayed in the page).

bianchini88:
Updates on documentation of FAIRsharing IDs.

Closes #1553


bedroesb (approved):

bedroesb:
Thanks! You can merge it if you want.

PR #1647, opened 2025-06-16

bianchini88:
Full name of resource


bedroesb:
image


I took the out, to prevent repetition. But you can revert it if needed

PR #1651, opened 2025-06-23
Update link for RDMP on storage

marikapop:
Update links on storage page

### DIFF
--- a/_data/tool_and_resource_list.yml
+++ b/_data/tool_and_resource_list.yml
@@ -1754,7 +1754,7 @@
   name: Research Data Management Platform (RDMP)
   registry:
     biotools: rdmp
-  url: https://www.dundee.ac.uk/hic/data-team/researchdatamanagementplatform/
+  url: https://www.dundee.ac.uk/hic/data-service/research-data-management-platform
 - description: Machine actionable DMPs.
   id: research-management-plan
   name: Research Management Plan


_data/tool_and_resource_list.yml
```diff
@@ -1754,7 +1754,7 @@
   name: Research Data Management Platform (RDMP)
```
Nazeeefa, line 1754: I think the full name should be "Research Data Management Platform (RDMP) at Health Informatics Centre, University of Dundee"

Nazeeefa (changes requested):

_data/tool_and_resource_list.yml
bedroesb: That is a bit too long for a tool name, if it is not strictly needed imo. Can we not add this information to the description?

marikapop: I agree with Bert, this is too long for the tool name. Even current name is too long 
image
so I shortened it into RDMP and extended the description.

@bedroesb Bert can the tool id also be changed from  id: research-data-management-platform to id: rdmp? I do not know how to check if there will be no conflicts with such id (yaml file does not have such id yet but what about other places?).

bedroesb: make sure to open the pull request in the vscode editor so you can search for all appearances of `research-data-management-platform`. Those will need to be updated. To open it in the editor, go to the PR and press . (dot)

bedroesb: I can also help off course. btw I think with @bianchini88 we kinda decided to spell out full tool names with the acronym in parentheses.

marikapop: > I can also help off course. btw I think with @bianchini88 we kinda decided to spell out full tool names with the acronym in parentheses.

Do you mean that id should be `research-data-management-platform_(rdmp)`?

bedroesb: @marikapop oh no, I ment the name, not the id. it is a good idea to keep the id short (with the acronym) So full support on that one!

_data/tool_and_resource_list.yml
```diff
@@ -1749,12 +1749,12 @@
   id: research-data-management-organiser
   name: Research Data Management Organiser
   url: https://rdmorganiser.github.io/en
-- description: Data management platform for automated loading, storage, linkage and provision of data sets
+- description: Research Data Management Platform (RDMP) at Health Informatics Centre, University of Dundee for automated loading, storage, linkage and provision of data sets
   id: research-data-management-platform
```
bedroesb:
```suggestion
  id: rdmp
```

And off course also changing the tool mentions/snippets themseleves

_data/tool_and_resource_list.yml
```diff
@@ -1749,12 +1749,12 @@
...
-  name: Research Data Management Platform (RDMP)
+  name: RDMP
```
bedroesb:
```suggestion
  name: Research Data Management Platform (RDMP)
```

@bianchini88 I personally like the combination full name + acronym, like we have been doing for many other tools. THe tool id itself is better the acronym imo

bedroesb (approved):

bedroesb:
@Nazeeefa Since these are link fixes and the tool names currently are in line with our style guide/similar to other tools in RDMkit, I will go ahead and merge this. If you think something still has to be changed, do not hesitate to reply to this thread.

PR #1652, opened 2025-06-23
fixed links  CODE_OF_CONDUCT.md

diana-pilvar:


_data/events.yml
Nazeeefa, line 1: For entry "ELIXIR CONVERGE - RDMkit Virtual Hackathon", it also says "Sign up here" that needs to be replaced with "Registration has ended" just as similar to "Contentathon" entry.

PR #1689, opened 2025-06-24
Events do not show headings

bedroesb:


bedroesb:
Then upcoming events will be available, there will be an extra heading.

korbinib (approved):

PR #1696, opened 2025-07-07

sneumann:
- Moved Argos up (adhere to alphabetical sorting
- Removed duplicated identical "DMP Canvas Generator" entry


diana-pilvar:
> To the reviewers: consider whether "DMP Canvas Generator" should stay here or whether it should be moved to https://rdmkit.elixir-europe.org/ch_resources

I agree, this should be on CH resources as I can't log in even with orcid.

diana-pilvar (approved):

PR #1699, opened 2025-07-18
Prevent duplicates in the link checker

bedroesb:


korbinib (approved):
we could also use`lycheeVersion: latest`

bedroesb:
Finally the exclude works!

PR #1733, opened 2025-11-05
Fix link checker run

bedroesb:


bedroesb:
@bianchini88 THis fix works but I doesn't pickup all errors anymore: https://github.com/elixir-europe/rdmkit/actions/runs/19095586211

Issue #1717, opened 2025-08-27
Update citation.cff file with latest publication information


bedroesb:
https://www.cell.com/patterns/fulltext/S2666-3899(25)00193-X

bedroesb:
And also update the about!

bedroesb:
Update the [README](README.md) to have both the Citing the repo (content) and the whole resource (publication), similar in how we will do it in the about. 

Extra ToDo: Change ELIXIR author in citation.cff to RDMkit Community

PR #1744, opened 2025-11-21
Citation.cff

korbinib:
Revise citation information for RDMkit
Updated citation details for RDMkit including authors, version, and preferred citation format.
Closes https://github.com/elixir-europe/rdmkit/issues/1717

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,19 +1,20 @@
 source 'https://rubygems.org'
 
 gem 'elixir-toolkit-theme-plugins', '~> 0.1.10'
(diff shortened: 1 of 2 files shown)


bedroesb:
@korbinib I wonder if the preferred citation is the paper of our current citation line 🤔

CITATION.cff
```diff
@@ -1,16 +1,81 @@
 cff-version: 1.2.0
```
diana-pilvar: should we update to 1.2.1?

PR #1750, opened 2025-12-04
Create Federated_Learning

smza:
Started a new page for Federated Learning. Author name: Jorge Silva

### DIFF
--- a/pages/your_tasks/Federated_Learning
+++ b/pages/your_tasks/Federated_Learning
@@ -1,0 +1,47 @@
+---
+title: Federated Learning Best Practices
+description: <!---REPLACE THIS with a one sentence description of the page--->
+contributors: [Jorge Silva]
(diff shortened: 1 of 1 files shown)


pages/your_tasks/Federated_Learning
bedroesb, line 1: no capitals in the filename

pages/your_tasks/Federated_Learning
```diff
@@ -0,0 +1,47 @@
+---
+title: Federated Learning Best Practices
```
bedroesb, line 2: Maybe a temporary working title, but I would try to follow our current title structure. What is the task that a researcher/data steward has to tackle, in the context of data management. I think we should also be careful this does not talk to much about software.

smza: @bedroesb yes it is a temporary working title. I have advised about the scope and context of the Your Task pages to the author. I'll be working with them to get it right.

PR #1759, opened 2025-12-12

bianchini88:
Added missing FAIRsharing ID (why was this not picked up by the tool, btw?)

### DIFF
--- a/_data/tool_and_resource_list.yml
+++ b/_data/tool_and_resource_list.yml
@@ -2686,7 +2686,7 @@
   url: https://loinc.org
 - description: Terminology standard consisting of codes, terms, synonyms, and definitions used in clinical documentation and reporting.
   id: snomed-ct
-  name: Systematized Nomenclature of Medicine - Clinical Terms (SNOMED-CT)
+  name: Systematized Nomenclature of Medicine Clinical Terms (SNOMED CT)
   registry:
     fairsharing: d88s6e
   url: https://www.snomed.org/what-is-snomed-ct


bianchini88:
@bedroesb Now that you brought this up, I actually think "SNOMED CT" is the correct one. See https://www.snomed.org/what-is-snomed-ct

_data/tool_and_resource_list.yml
```diff
@@ -2684,10 +2684,12 @@
   id: loinc
   name: Logical Observation Identifiers Names and Codes (LOINC)
   url: https://loinc.org
-- description: Terminology standard consisting of codes, terms, synonyms and definitions used in clinical documentation and reporting.
+- description: Terminology standard consisting of codes, terms, synonyms, and definitions used in clinical documentation and reporting.
   id: snomed-ct
   name: Systematized Nomenclature of Medicine - Clinical Terms (SNOMED-CT)
```
bianchini88:
```suggestion
  name: Systematized Nomenclature of Medicine Clinical Terms (SNOMED CT)
```

bedroesb: hmmm taking those spaces does not look good imo. Is there an example on their website? I can't seem to find it

bedroesb: hmmm taking those spaces does not look good imo. Is there an example on their website? I can't seem to find it

bianchini88: See comment below, it should be a en dash.

bianchini88:
And [here](https://www.england.nhs.uk/digitaltechnology/digital-primary-care/snomed-ct/) it is without dashes, which is a bit cleaner.

_data/tool_and_resource_list.yml
bianchini88: Updated, as it often appears also without dashes.

PR #1811, opened 2026-03-13
Add AgriSchemas to tool and resource list

Arnedeklerk:
Added AgriSchemas entry with relevant details.
Updated KnetMiner
Added some biotools links


bedroesb:
@Arnedeklerk this looks great! Could you make these changes here: https://github.com/Arnedeklerk/rdmkit/blob/patch-1/_data/tool_and_resource_list.yml This way they will be part of the same pull request and the workflows will not fail anymore! We can than close this pull request.

PR #1833, opened 2026-04-24
Implement retry mechanism for registry lookup

korbinib:
Added retry logic for registry lookup in tools validator script.


bianchini88:
Moving to draft. 

See results [here](https://github.com/elixir-europe/rdmkit/actions/runs/24890502836/job/72881287739). Failure is consistent.

PR #1847, opened 2026-06-10
Use fuzzy search for TeSS lookup

bedroesb:


bedroesb:
If we want to annotate TeSS entries with Bio.tools we might want to have exact matches though ...

_data/tool_and_resource_list.yml
```diff
@@ -120,6 +121,7 @@
   name: Privacy Impact Assessment Tool (PIA)
   registry:
     biotools: pia
```
bianchini88, line 123: https://bio.tools/pia this looks something else entirely...

_data/tool_and_resource_list.yml
```diff
@@ -120,6 +121,7 @@
...
+    tess: Privacy Impact Assessment Tool
```
bianchini88, line 124: The TeSS entry is this one https://tess.elixir-europe.org/materials/data-protection-and-security-aspects-of-running-simulations-on-personal-data-with-hpc which seems to be something else.

_data/tool_and_resource_list.yml
```diff
@@ -302,6 +304,7 @@
   name: Biodiversity Information Standards (TDWG)
```
bianchini88, line 304: This is actually an organisation, so it should not be here in the first place. The actual standards this organisation is maintaining should be here, i.e. https://www.tdwg.org/standards/

_data/tool_and_resource_list.yml
```diff
@@ -302,6 +304,7 @@
...
   registry:
     fairsharing: '826786'
+    tess: Biodiversity Information Standards
```
bianchini88, line 307: Many results, difficult to assess whether they are relevant, considering this should not be an entry in the first place.

_data/tool_and_resource_list.yml
```diff
@@ -408,6 +411,7 @@
   name: Cancer Genome Interpreter
   registry:
     biotools: cgi
+    tess: Cancer Genome Interpreter
```
bianchini88, line 414: Results might not be relevant, e.g. https://tess.elixir-europe.org/materials/webinar-variant-interpretation-from-the-clinic-to-the-lab-and-back-again

bianchini88 (changes requested):
I will not finish this review. Halfway through, I realised that we are introducing too much spurious content. So I would be against merging this and moving back to the old system.

bedroesb:
Lets continue the dicsussion in #1841

PR #1849, opened 2026-06-11

### LINKED ISSUE #1848: Privacy Impact Assessment Tool (PIA) - the bio.tools ID is something else
This https://www.cnil.fr/en/open-source-pia-software-helps-carry-out-data-protection-impact-assessment and https://bio.tools/pia are clearly too different things, in spite of the acronym being the same.

bianchini88:
Removing bio.tools ID that was not matching the entry

### DIFF
--- a/_data/tool_and_resource_list.yml
+++ b/_data/tool_and_resource_list.yml
@@ -118,8 +118,8 @@
 - description: The open source PIA software helps to carry out data protection impact assessment
   id: pia
   name: Privacy Impact Assessment Tool (PIA)
-    registry:
-      biotools: NA
+  registry:
+    biotools: NA
   url: https://www.cnil.fr/en/open-source-pia-software-helps-carry-out-data-protection-impact-assessment
 - description: The Japanese Genotype-phenotype Archive (JGA) is a service for permanent archiving and sharing of all types of individual-level genetic and de-identified phenotypic data resulting from biomedical research projects.
   id: jga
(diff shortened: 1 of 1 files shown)


bedroesb:
@bianchini88 https://rdmkit.elixir-europe.org/tool_resource_update you have to use NA to prevent it gets re-added next Monday.

bianchini88:
@bedroesb I thought the acronyms were excluded from the matching, and acronyms here are the only thing matching. I mean, one is called "Privacy Impact Assessment" and the other "Protein Inference Algorithms".

bedroesb:
@bianchini88 with bio.tools the acronym is most of the time the identifier, youa re right that we don't do this for TeSS because to many false positives. 

https://github.com/elixir-europe/rdmkit/blob/master/_data/tool_and_resource_list.yml#L696 

Is an example : DNA Data Bank of Japan (DDBJ) -> https://bio.tools/ddbj

bianchini88:
@bedroesb I see, but in this case the full name is "PIA - Protein Inference Algorithms". Do you think we will end up getting a match?

bedroesb:
With a hyphen not, like it is now, Privacy Impact Assessment Tool (PIA), we will.

_data/tool_and_resource_list.yml
```diff
@@ -118,8 +118,6 @@
 - description: The open source PIA software helps to carry out data protection impact assessment
   id: pia
   name: Privacy Impact Assessment Tool (PIA)
```
bedroesb, line 120:
```suggestion
  name: Privacy Impact Assessment Tool (PIA)
    registry:
      biotools: NA
```

bianchini88:
@bedroesb good to go?

bedroesb (approved):

PR #1850, opened 2026-06-11
Update CODEOWNERS

bianchini88:
Adding myself to "tools and resources", as I do a lot of curation and annotation on that file.


bedroesb (approved):

bedroesb:
Great!

PR #1851, opened 2026-06-11
Starting with consistent use of (f)EGA

### LINKED ISSUE #1605: Federated EGA is not used as tools snippet in RDMkit
The tool is untagged:

Image

Yet there is lots of potential:

Image

bedroesb:
This should eventually close #1605 

Interesting, the current FEGA nodes are listed as instances of EGA, not FEGA.

image

We can continue in the issue the discussion.


bianchini88:
This is a bit complicated, as not necessarily all the FEGA nodes use the same technology stack (i.e. they are not necessarily instances of the same entity). One should probably also start introducing the nomenclature CEGA (Central EGA), and specify a bit better that the FEGA nodes are hosting the data, whilst the EGA portal exposes metadata. Tagging @korbinib, as he probably has a better understanding than I have. Btw, FEGA is much larger than what we represent on RDMkit today, see e.g. https://ega-archive.org/about/projects-and-funders/federated-ega/

korbinib:
This is also a bit of a question of the target audience: For a researcher that wants to submit their data it might be irrelevant which software stack their FEGA node is running on. It does matter of course if you want to deploy one yourself. But then also the legal work might be the more challenging part in the first place.

bianchini88:
Fully agree with @korbinib. One also has to mention that the local node and the central portal will have to be used for deposition (one for data and the other for metadata), whilst only EGA is used for data discovery. In this sense, it is not correct to mark FEGA nodes as "instances" of EGA, as national nodes do not provide (by construction) the full set of functionalities. FEGA nodes could be considered instances of FEGA (although not necessarily based on the same stack) because they all play the same role in the various countries.

korbinib (approved):

bedroesb:
I guess this is ready to go!

PR #1859, opened 2026-07-24
Update tools_validator.py

bianchini88:
Adding a delay to account for rate limit on FAIRsharing's API

### DIFF
--- a/var/tools_validator.py
+++ b/var/tools_validator.py
@@ -212,7 +212,7 @@ with open(yaml_path, 'r') as read_obj:
                             registry['fairsharing'] = check_fairsharing
                 if 'fairsharing' in registry.keys() and not registry['fairsharing']:
                     del registry['fairsharing']
-                time.sleep(2) # to avoid issues with rate limit
+                time.sleep(5) # to avoid issues with rate limit
 
                 # Add populated registry dict to the main list
                 if registry:
--- a/var/tools_validator.py
+++ b/var/tools_validator.py
@@ -9,6 +9,7 @@ from requests.adapters import HTTPAdapter
 from urllib3.util.retry import Retry
 import frontmatter
 from urllib.parse import urlparse
+import time
 
 class NullRepresenter:
     def __init__(self):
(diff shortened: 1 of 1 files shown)


var/tools_validator.py
```diff
@@ -211,6 +212,7 @@ def remove_prefix(s, prefix):
                             registry['fairsharing'] = check_fairsharing
                 if 'fairsharing' in registry.keys() and not registry['fairsharing']:
                     del registry['fairsharing']
+                time.sleep(1) # to avoid issues with rate limit
```
bianchini88:
```suggestion
                time.sleep(2) # to avoid issues with rate limit
```

bianchini88:
With a 5-second delay, things work until we get a 500 Internal Server Error. Rerunning.

var/tools_validator.py
```diff
@@ -211,6 +212,7 @@ def remove_prefix(s, prefix):
                             registry['fairsharing'] = check_fairsharing
                 if 'fairsharing' in registry.keys() and not registry['fairsharing']:
                     del registry['fairsharing']
+                time.sleep(5) # to avoid issues with rate limit
```
bedroesb, line 215: ooof I am not a fan of this solution. This will increase the general time way to much no? I assume we can do something more clever on the requests side, or keeping track of the number of requests we've already done.

bedroesb:
@bianchini88 You have better contacts with FAIRsharing, can this not be solved on their end? I understand rate limits for unauthenticated requests, but we are authenticated, and only doing around 1000 requests over 10min. This does not feel like a hard ask. I am not comfortable putting sleeps like this in the code

bianchini88:
Crashed twice at European Joint Programme on Rare Diseases Metadata Model, which should get https://fairsharing.org/FAIRsharing.3e0038. Not sure about the issue.

bianchini88:
Hi @bedroesb. This was meant as a temporary fix for the summer, but it does not look like it is working anyway (and I am off in 15 minutes to catch a flight). FAIRsharing recommends that we transition to the GraphQL API and the 'advancedSearchFast' query (https://api.fairsharing.org/schema/query.doc.html).

PR #1870, opened 2026-09-01
Add AGENTS.md instructions for AI coding assistants

nilshoffmann:
Document the repository conventions that AI assistants need in order to contribute usable pull requests: the page templates and front matter, the distinction between in-text links (which use the filename) and related_pages (which uses the page_id), how the {% tool %} tag expands, and the rules for adding entries to the tool table, contributors and the bibliography.

CLAUDE.md only imports AGENTS.md, because Claude Code reads CLAUDE.md while other assistants follow the AGENTS.md convention.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,309 @@
+# AGENTS.md
+
+Instructions for AI coding assistants working in the RDMkit repository. This file is
+assistant-agnostic: tools that follow the `AGENTS.md` convention read it directly, and
+you can point any other assistant at it manually.
+
+Claude Code is an exception — it reads `CLAUDE.md`, never `AGENTS.md`. The repository
+root therefore has a one-line `CLAUDE.md` containing `@AGENTS.md`, which imports this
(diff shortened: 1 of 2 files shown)


bedroesb:
@nilshoffmann Thanks a lot for start the discussion on this one. I personally am not a fan in copying the content of our style guide/editorial guide/etc into a new file, because that means we have to maintain 2 files for the same content. Could you transfer this in a GitHub action that melts the relevant files together into an agent file on the fly?

nilshoffmann:
@bedroesb Thanks for your feedback. I would rather than introducing a GitHub action for this just reference the original guides in pages/contribute/, which most AIs & agents should be able to resolve. This would locate anything agent specific in the AGENTS.md and keep all content relevant to both humans and agents in the established locations. I would however also add a bit of documentation for humans that Claude stumbled over, may be beneficial for either:

- in-text links use the file name, not the page_id (markdown cheat sheet and editorial board guide)
- {% tool %} prints the tool's full name, so don't spell it out around the tag (tool_resource_update).

